Issue #9, opened 2017-09-22
Support the Sphinx latex builder


Mankro:
I cannot compile RST chapter content to a PDF (run `make latexpdf` in an RST course directory) if the RST code uses any aplus directives, such as `aplusmeta` in round index files. If I remove the aplus directives from the RST code, then `make latexpdf` works again. Note that `a-plus-rst-tools/aplus_setup.py` calls `app.add_node` with visitor functions only for the `html` builder. According to http://www.sphinx-doc.org/en/stable/extdev/appapi.html#sphinx.application.Sphinx.add_node it is then expected that the `latex` builder is unable to process the directive. It could be enough that the output of the directive is empty when the latex builder is used, basically ignoring the directives completely. Currently, the latex builder ignores the files with these aplus directives with a misleading error message: `index.rst: WARNING: toctree contains ref to nonexisting file 'r02/index'` (assume that `r02/index uses the aplusmeta directive). The output PDF does not include any r02 content in this case.

I have tried to define empty visitor functions for latex in the function calls `app.add_node` in the file `aplus_setup.py`, but it did not change anything. I think it should work, but something still gets wrong. Thus unfortunately, I have no working solution.

Mankro:
The merged pull request #10 did not fix the whole thing. It is possible to compile the PDF now (instead of crashing to errors during the compilation), but the aplus RST directives (such as questionnaire) do not behave well in PDFs. The PR added a comment in https://github.com/Mankro/a-plus-rst-tools/blob/37c2b1e24431b5d3347fbbf1a8f851456b1e77ef/aplus_setup.py#L42 which says the following:

Note: even though these latex visitor functions do nothing, the exercise directives nonetheless output some text to the Latex/PDF output. The text is even often duplicated multiple times. This is possibly caused by broken aplus_nodes classes that interfere with the builder output even though only the visitor functions should do that.

PR #10, opened 2017-09-27
Support the Sphinx latex builder when A+ RST directives are used

Mankro:
Related issue #9 

Normally you support different builders by defining visit methods for them for each node when `setup` calls `app.add_node`. However, it was not enough with `a-plus-rst-tools` due to other problems.

`aplus_nodes.html` and `.aplusmeta` inherit `docutils.nodes.Element`, which has a constructor
that expects many keyword arguments. The classes in `aplus_nodes` used to override
the constructor in such a way that it did not accept those other arguments expected by
the base class and this caused a crash, for example, `TypeError: __init__() got an unexpected keyword argument 'dupnames'`. However, it seems that this issue does not affect the Sphinx `html` builder while it definetely affects the `latex` builder.

With the new code in the pull request, the node class constructors do not crash if they receive more keyword arguments that are intended for the superclass constructor. There is an overridden copy method as well because at least with the `latex` builder, Sphinx runs a deep copy 
of the document tree and it uses a monkey-patched copy method for the `docutils.nodes.Element` class (see `sphinx.util.nodes._new_copy` [source](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/nodes.py#L365)). The constructor is called like `newnode = self.__class__(self.rawsource, **self.attributes)`, which could imply that it is safer to define the `aplus_nodes` constructors with prototypes that start with the same positional parameters as the superclass's constructor.
Alternatively, perhaps the `aplus_nodes` classes should override the `copy` method, which is what I did. Without the overridden copy methods, the constructors would be called with incorrect arguments when the `latex` builder is run.

(The `tagname` variable should be defined as a class variable according to the API docs. That is a problem with `aplus_nodes.html` since the tagname actually varies, but at least the `aplusmeta` does not need to define the `tagname` at all since the default value taken from the class name is good enough.)

Some of the course configuration YAML generation has been bundled into the HTML visit methods, thus only `make html` builds the config YAML properly. It is not worthwhile to refactor the old code right now.

Exercise directives still output something to the Latex PDF even though empty visitor functions are used. The aplus_nodes classes are probably broken in that sense since only the visitor function should affect the builder output.

### DIFF
--- a/aplus_nodes.py
+++ b/aplus_nodes.py
@@ -22,13 +22,15 @@ class html(nodes.General, nodes.Element):
         Captures rendered HTML from children nodes that were requested
         to store HTML.
     '''
-
-    def __init__(self, tagname, attributes={}, no_write=False, skip_html=False):
+    
+    def __init__(self, tagname, attributes={}, no_write=False, skip_html=False, *children, **other_attributes):
         ''' Constructor: no_write option removes node from final document after configuration data is processed. '''
         self.tagname = tagname
         self.no_write = no_write
         self.skip_html = skip_html
-        super(html, self).__init__(rawsource=u"", **attributes)
+        attrs = other_attributes.copy()
+        attrs.update(attributes) # attributes dict gets prioritized if the same key exists in both dicts
+        super(html, self).__init__(u"", *children, **attrs)
 
     def write_yaml(self, env, name, data_dict, data_type=None):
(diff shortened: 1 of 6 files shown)


teemulehtinen:
Yes. It is deliberately broken as I don't like rst too much :) There are actually some historical versions that did support latex output which I then decided to break to avoid logical nodes in favor of fewer presentation nodes.

PR #13, opened 2018-01-24
Raise an error if a questionnaire contains an unknown compare_method

jrp6:
Fix silent failure in case of a typo by raising a SphinxError if a questionnaire contains an undefined compare_method (e.g. `stirng` instead of `string` or `regex` instead of `regexp`) or a modifier (e.g. `ignorecasd` instead of `ignorecase`).

Note that if an exercise has a config in a separate YAML file, the compare_method specified in the rst file has no effect and is not checked.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -412,6 +412,7 @@ class FreeText(QuestionMixin, Directive):
 
         # Add configuration.
         if len(self.arguments) > 1:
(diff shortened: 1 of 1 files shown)


raphendyr:
Thanks. For future reference, this will be replaced with json-schema validation at some point. At least, if schema validation is good enough solution (might not be due to complicated data format).

PR #14, opened 2018-02-08
make MathJax optional

jsorva:
MathJax was included in all HTML output. Now it is still on by default
but easily disabled with a Sphinx theme option.


Mankro:
the MathJax `<script>` element does not have the `data-aplus` attribute though, so it is not used in A+ anyway. It applies when the HTML file is opened in the `_build` directory ("mock frame").

jsorva:
@Mankro Fair enough. It interfered slightly with my local testing outside of A+, though, hence the change.

raphendyr:
This is logical change in any case.

PR #15, opened 2018-04-17
Fixed deprecated Directive import

Donaldi3:
sphinx.util.compat has been deprecated and shouldn't be used anymore


raphendyr:
I'll look into this at some point, when I reread all of the sphinx directives and refactor this to a tool instead of extension. But thanks. I now know where to look for the Directive class to read the code.

raphendyr:
Can you create PR from the last 2 commits.

PR #19, opened 2018-09-04
Module key based on dir structure

piehei:
Hello!

Problem:
If your module hierarchy is:
- index
- modules/01/index
- modules/02/index
- modules/nn/index

all module **_keys_** in the generated yaml file will be "modules". This of course conflicts with everything, and A+ happily merges all the material into a single round which has the configuration of the last-defined module. In other words, current implementation only works for modules that are placed in the root in single directories, not in directories inside directories.

This PR fixes this problem.

Current behaviour:
* module key comes from the root dir name of that module

New behaviour:
* module key comes from the whole dir hierarchy of that module, all occurences of "/" are replaced with "_"

### DIFF
--- a/toc_config.py
+++ b/toc_config.py
@@ -183,7 +183,11 @@ def make_index(app, root):
         close_src = meta.get('close-time', title_date_match.group(1) if title_date_match else course_close)
         late_src = meta.get('late-time', course_late)
         module = {
-            u'key': docname.split(u'/')[0],
+            # modules01/index -> modules01
+            # modules/01/index -> modules_01
+            # modules/01/n/index -> modules_01_n
+            # ...
+            u'key': docname if u'/' not in docname else u'_'.join(docname.split(u'/')[:-1]),
             u'status': status,
             u'name': title,
             u'children': [],


piehei:
Something to note: previously generated keys should stay unchanged.

raphendyr:
Should work and looks better than original. Needs to be tested that it doesn't break existing courses.

piehei:
One thing... This of course breaks, if someone has both a module dir module_01 AND module/01 since they will end up with identical keys. No sensible person should lay out the directories in that way, of course :-)

Duplicate keys should be checked in a future release or something.

piehei:
We're starting to build multiple new courses for the Spring period and would *LOVE* to have this functionality.

Could this be merged? We don't want to have a fork with a set of custom patches. :)

toc_config.py
```diff
@@ -183,7 +183,11 @@ def parse_chapter(docname, doc, parent):
         close_src = meta.get('close-time', title_date_match.group(1) if title_date_match else course_close)
         late_src = meta.get('late-time', course_late)
         module = {
-            u'key': docname.split(u'/')[0],
+            # modules01/index -> modules01
+            # modules/01/index -> modules_01
+            # modules/01/n/index -> modules_01_n
+            # ...
+            u'key': docname if u'/' not in docname else u'_'.join(docname.split(u'/')[:-1]),
```
raphendyr: Old behavior is changed in the code: `docname.split(u'/')[0]` -> `docname`

PR #20, opened 2018-09-07
Fix exercise difficulty in the hacky traky branch

Mankro:
The traky branch is dirty, but a course is still depending on it. This PR adds the exercise difficulty configuration to the branch.


raphendyr:
I close my eyes and press merge... Is that ok?..

PR #23, opened 2018-11-07
Directive for ExerciseCollection

Donaldi3:
Directive works.
Couldn't get error reporting working properly, so I used assert for severe problems.


raphendyr:
Can you fix the conflict and also remove carriage returns from your new file.

raphendyr:
And also squash these two to one.

PR #28, opened 2019-01-17
Fix image URLs in exercise descriptions for A+

Mankro:
The A+ frontend must convert relative URLs in the course materials
to absolute since the URLs must refer to the backend server,
not to A+, when the material is viewed through A+.
 source URLs in exercise descriptions were not handled
correctly in the A+ frontend. They worked in chapter contents but
not in exercise descriptions because the service URLs for chapters
are based on the MOOC grader static URL while the exercise service
URLs are different than the static URL.

This fix adds data-aplus-path attribute to the img and similar
elements that link to a local file. This is done only in exercises,
not in the chapter contents. The attribute allows A+ to set
the image URL correctly.

### DIFF
--- a/lib/html_tools.py
+++ b/lib/html_tools.py
@@ -31,25 +31,27 @@ def rewrite_file_links(path, root, chapter_dirs, static_host):
         static_host,
         chapter_dirs,
         u'data-aplus-chapter="yes" ',
+        u'data-aplus-path="/static/{course}" ',
     )
     _write_file(path, content)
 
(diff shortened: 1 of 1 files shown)


Mankro:
With this fix, one may use the `.. image::` directive normally inside the `.. questionnaire::` directive. Furthermore, relative links such as 
```
`astar.zip <../_static/astar.zip>`_
```
work inside the `questionnaire`.

PR #31, opened 2019-01-27
Add support for more LTI options

piehei:
Code should be completely self-explanatory.
A-plus already parses this from the course config JSON so this only adds the support for this toolset.


raphendyr:
Yep, you could check existense of those options using the same method as for `lti` now. I.e. `if 'lti' in self.options:`.

pietariheinovincit:
It would be super cool if someone could finish this for me – I see some development happening and new faces here :)

Mankro:
@alalav1 could check this out after finishing current tasks.

Mankro:
PR #37 superseded this PR so this PR was closed. The merged commit has Pietari as a co-author in order to give some credit for this original PR.

PR #32, opened 2019-01-28
Cleaned up ExerciseCollection configuration

Donaldi3:
Renamed collection_url and collection_category to target_url and target_category for more intuitive usage.
Removed collection_course option, because it was removed from a-plus' side (wasn't really necessary and harder to use)

Removed carriage returns from directives/exercisecollection.py

### DIFF
--- a/directives/exercisecollection.py
+++ b/directives/exercisecollection.py
@@ -11,10 +11,10 @@ from directives.abstract_exercise import AbstractExercise
 
 class ExerciseCollection(AbstractExercise):
     has_content = True
+    required_arguments = 1
     option_spec = {
-        'collection_course': directives.unchanged,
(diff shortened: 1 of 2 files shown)


raphendyr:
Why there is so many "changes" for the directive file? Line endings?

Donaldi3:
Yes. Removing the carriage returns.

raphendyr:
These are technically backwards incompatible changes.. :disappointed: 
This time, only this time, I expect no one else is using this feature and a-plus-rst-tools is enough prototype for breaking semantic versioning rules.

With Arial, we will take better care that we don't do similar things.. In addition, all new features should be first implemented as separate extensions and only merged to core when they are mature enough. More about that when we get there :slightly_smiling_face:

PR #34, opened 2019-05-20
Point of interest additional options

roosapi:
Add height, columns, and background image options for point of interest.

Swap title and id so that title becomes an argument and id an option (random id generated if one not given).

### DIFF
--- a/directives/point_of_interest.py
+++ b/directives/point_of_interest.py
@@ -2,17 +2,25 @@
 '''
 Directive for creating "point of interest" summary block.
 
-.. point-of-interest:: name (unique id within the document)
-    :title: optional title text
+.. point-of-interest:: title text
+    :id: unique id within the document
     :previous: name of previous point-of-interest
     :next: name of next point-of-interest
     :hidden: (if this flag is present, the content of this poi is hidden by default)
     :class: any additional css classes
+    :height: optional fixed height for content
+    :bgimage: path to background image
+    :columns: relative widths of poi content columns (e.e. 2 3 3)
 
     Content of point-of-interest here
 
+    ::newcol
+
+    Using ::newcol separated by line breaks starts a new column
+
 '''
-import os.path
+import os.path, random, string
+from math import floor
 from docutils.parsers.rst import Directive, directives
 from docutils import nodes
 from sphinx.errors import SphinxError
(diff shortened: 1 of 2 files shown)


Mankro:
The changes seem fine to me. I assume you have tested it thoroughly so that it works correctly.

General comments:

* The API of the directive changes here, so the material creators must be aware. This is even more important when this directive is used in two courses whose teachers will change now and they don't know these technical details very well.
* `STATIC_CONTENT_HOST` environment variable is required because we did not find a better way to fix the relative background image URL. Absolute URLs work in A+ and the environment variable enables building an absolute URL with the domain part.

raphendyr:
As the API changes, there are two options:

a) add new directive, e.g. `.. point-of-interest2::`

b) The code needs to handle both interfaces (old and the new). So, the code could check that if there is a `:title:`, then the argument for the directive is the id. If there is no `:title:`, then the directive argument is the title and `:id:` is used or one is generated. If there is `:title:` and `:id:`, then an exception is raised.

Mankro:
Indeed, handling backwards-compatibility would be important since people hate useless API changes, including myself. Raphendyr's idea sounds very reasonable and easy to implement as well.

roosapi:
Fixed so that the old API is supported according to option B and squashed commits

directives/point_of_interest.py
```diff
@@ -32,50 +40,106 @@ class PointOfInterest(Directive):
...
-        title_text = ''
-        if 'title' in  self.options:
+
+        if 'title' in self.options:
+            if 'id' in self.options:
+                raise SphinxError(u'Point of interest options can\'t contain both "title" and "id"; one of them should be provided as an argument instead')
+            name = self.arguments[0]
             title_text = self.options['title']
+        else:
+            name = self.options.get('id', ''.join(random.choice(string.ascii_lowercase + string.digits) for i in range(6)))
```
raphendyr: `.get('id') or random_string_generation` would be bit faster. Currently, the random string is always generated. Bit better looking code would be:

```
name = self.options.get('id')
if not name:
  choices = string.ascii_lowercase + string.digits
  name = ''.join(random.choice(choices) for i in range(6))
```

p.s. there is still change that the name is not unique.
p.s.s. `choices` is optimization, which is not done automatically by Python

directives/point_of_interest.py
```diff
@@ -32,50 +40,106 @@ class PointOfInterest(Directive):
...
+            title_text = self.arguments[0]
```
raphendyr: Does this handle spaces in the string? e.g.

```
.. point-of-interest:: My long title with multiple words
```

directives/point_of_interest.py
roosapi: I'll change that and add a check to make sure the id is unique. Not quite sure what you mean with the last one though?

directives/point_of_interest.py
roosapi: Yes, spaces in the title are fine.

raphendyr: Good. I didn't remember how the arguments are split by sphinx.

directives/point_of_interest.py
raphendyr: Just an information why I wrote the example like that. Real programming language would do the optimisation, but python can't.

roosapi: Ah ok, I see. Changed this and in the process noticed that poi info was not correctly cleared from the environment so added a fix for that as well.

directives/point_of_interest.py
```diff
@@ -32,50 +40,114 @@ class PointOfInterest(Directive):
...
+
+        if 'title' in self.options:
+            if 'id' in self.options:
+                raise SphinxError(u'Point of interest options can\'t contain both "title" and "id"; one of them should be provided as an argument instead')
+            name = self.arguments[0]
             title_text = self.options['title']
+        else:
+            name = self.options.get('id')
+            if not name:
+              choices = string.ascii_lowercase + string.digits
```
Mankro: Indentation of only 2 spaces when it should be 4?

raphendyr:
Looks good. I didn't review the html generation very well, but I presume old versions are not broken too much.

directives/point_of_interest.py
roosapi: Hups, fixed!

Mankro:
Are we ready to merge?

raphendyr:
Fine by me.

PR #36, opened 2019-07-04
option to set exercise status manually

alalav1:
First try on this. Some of old code still commented, so still needs to be modified a little. Trello issue: https://trello.com/c/pj4iNdQa/34-exercise-status-must-not-be-hardcoded-in-rst-tools

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ The questionnaire directive accepts the following options:
 * `pick_randomly`: integer. Set the pick_randomly setting for the quiz
   (select N questions randomly on each load)
 * `category`: exercise category
+* `status`: exercise status (default "unlisted"). See available [statuses](#list-of-exercise-statuses).
 
 The contents of the questionnaire directive define the questions and possible
 instructions to students.
@@ -353,6 +354,7 @@ It accepts the following options:
 * `class`: CSS class(es)
 * `title`: exercise title
 * `category`: exercise category (default "submit")
+* `status`: exercise status (default "unlisted"). See available [statuses](#list-of-exercise-statuses).
 * `ajax`: If set, the A+ chapter does not attach any JavaScript event listeners
   to the exercise and the exercise JS may control the submission itself.
   See [the chapter content documentation](https://github.com/Aalto-LeTech/a-plus/blob/master/doc/CONTENT.md)
@@ -475,6 +477,7 @@ More active element examples can be found at https://version.aalto.fi/gitlab/pii
                     active elements)
       :scale-size: no value; if this option is present, the output element
                    height will scale to match content that has a defined height
+      :status: exercise status (default "unlisted"). See available [statuses](#list-of-exercise-statuses).
 
 ### 8. Hidden block
 
@@ -631,4 +634,13 @@ the mp4 or webm format. The id argument is the filename without the extension.
   :frame-height: 500
   :frame-width: 850
 ```
+### List of exercise statuses
 
+There are 6 possible statuses for exercises:
+
+* ready: Visible exercise listed in table of contents.
+* unlisted (default): Unlisted in table of contents, otherwise same as ready.
+* hidden: Hidden from non course staff.
+* enrollment: Questions for students when they enroll to a course.
+* enrollment_ext: Same as enrollment but for external students.
+* maintenance: Hides the exercise description and prevents submissions.
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -28,6 +28,7 @@ class Questionnaire(AbstractExercise):
         'points-to-pass': directives.nonnegative_int,
         'title': directives.unchanged,
         'category': directives.unchanged,
+        'status': directives.unchanged,
     }
 
     def run(self):
@@ -103,6 +104,7 @@ class Questionnaire(AbstractExercise):
             u'feedback': is_feedback,
             u'view_type': u'access.types.stdsync.createForm',
             u'title|i18n': translations.opt('feedback') if is_feedback else translations.opt('exercise', postfix=u" {}".format(key)),
+            u'status': self.options.get('status', 'unlisted'),
             u'fieldgroups': [{
                 u'title': '',
                 u'fields': (u'#!children', None),
--- a/directives/submit.py
+++ b/directives/submit.py
@@ -31,6 +31,7 @@ class SubmitForm(AbstractExercise):
         'radar_tokenizer': directives.unchanged,
         'radar_minimum_match_tokens': directives.unchanged,
         'category': directives.unchanged,
+        'status': directives.unchanged,
     }
 
     def run(self):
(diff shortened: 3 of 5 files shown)


toc_config.py
```diff
@@ -129,11 +129,12 @@ def parse_chapter(docname, doc, parent):
                     u'min_group_size': config.get(u'min_group_size', 1),
                     u'max_group_size': config.get(u'max_group_size', 1),
                     u'confirm_the_level': config.get(u'confirm_the_level', False),
+                    u'status': config.get(u'status', u'ready'),
                 }
             allow_assistant_viewing = config.get(u'allow_assistant_viewing', app.config.allow_assistant_viewing)
             allow_assistant_grading = config.get(u'allow_assistant_grading', app.config.allow_assistant_grading)
             exercise.update({
-                u'status': u'unlisted',
+                #u'status': u'unlisted',
```
raphendyr: Move the status here, so it applies to external exercises too: `u'status': config.get(u'status', u'unlisted'),`.
Also, use `unlisted` as a default (should be used for chapter exercises and RST generates those).

toc_config.py
```diff
@@ -163,9 +164,7 @@ def parse_chapter(docname, doc, parent):
         category = u'chapter'
         for name,hidden,child in traverse_tocs(app, doc):
             meta = first_meta(child)
-            status = u'hidden' if 'hidden' in meta else (
-                u'unlisted' if hidden else u'ready'
-            )
+            status = u'status'
```
raphendyr: This doesn't seem right. I have now idea where the `hidden` variable comes from and when there is `hidden` in meta. I recommend you look into this more and explain to me :)

alalav1: You are right, this is wrong. This defines chapter status, not exercise status. The task is to make submit exercises' statuses manually determinable, and chapter status should remain untouched.

toc_config.py
```diff
@@ -197,8 +198,8 @@ def parse_chapter(docname, doc, parent):
         title_date_match = title_date_re.match(title)
         meta = first_meta(doc)
         status = u'hidden' if 'hidden' in meta else (
-            u'unlisted' if hidden else u'ready'
-        )
+                u'unlisted' if hidden else u'ready'
```
raphendyr: You can use `git checkout -p upstream/master toc_config.py` and then say `y` to this part, to get the actual previous version. Of course, you need aalto-letech as an upstream repo (e.g. `git remote add upstream https://github.com/Aalto-LeTech/a-plus-rst-tools.git`).

toc_config.py
```diff
@@ -152,7 +153,7 @@ def parse_chapter(docname, doc, parent):
                 u'target_url': config[u'target_url'],
                 u'target_category': config[u'target_category'],
                 u'category': config[u'category'],
-                u'status': u'unlisted',
+                u'status': config[u'status'],
```
raphendyr: Should use `config.get(u'status', u'unlisted')`, similar to `exercise` type exercises. This block actually is for exercisecollection, which is implemented in #23.

directives/submit.py
```diff
@@ -101,6 +102,7 @@ def run(self):
             u'min_group_size': data.get('min_group_size', env.config.default_min_group_size),
             u'max_group_size': data.get('max_group_size', env.config.default_max_group_size),
             u'points_to_pass': self.options.get('points-to-pass', data.get('points_to_pass', 0)),
+            u'status': self.options.get('status', data.get('status','unlisted'))
```
raphendyr: Move this out of the `data.update` step and do something like: `data.setdefault('status', self.options.get('status', 'unlisted'))`

toc_config.py
```diff
@@ -129,11 +129,12 @@ def parse_chapter(docname, doc, parent):
                     u'min_group_size': config.get(u'min_group_size', 1),
                     u'max_group_size': config.get(u'max_group_size', 1),
                     u'confirm_the_level': config.get(u'confirm_the_level', False),
+
```
raphendyr: Why there is a extra new line here? :)

raphendyr:
This is readu for merging, but can you @alalav1 squash all the commits to a single one. Thanks.

git doc: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History (Squashing commits)

Mankro:
It looks like that only `submit` exercises are supported here. I think that `questionnaire` exercises should also be supported. It should be easy to add this to the questionnaire directive as well.

Mankro:
In addition, status should be set in active element outputs too (`directives/ae_output.py`).

directives/questionnaire.py
```diff
@@ -108,6 +109,7 @@ def run(self):
                 u'fields': (u'#!children', None),
             }],
         }
+        data.setdefault('status', self.options.get('status', 'unlisted'))
```
Mankro: In this case, `setdefault` always sets the key since the key `status` has never been set to the dict before this line. Thus, you could just set the key without calling setdefault. You can do it in the same statement that sets all those other keys.

Mankro: In the submit directive, the data dict may be read from a YAML file so the status key could be already there. That is why setdefault made sense in the submit directive.

Mankro:
By the way, update the README too in order to add the new options to the documentation.

Mankro:
Let's add the list of accepted status values to the README and also ensure in the code that only valid status values are accepted (otherwise, RST compilation should give an error to the user so that they must fix the status).

README.md
```diff
@@ -631,4 +634,11 @@ the mp4 or webm format. The id argument is the filename without the extension.
   :frame-height: 500
   :frame-width: 850
 ```
-
+###Available statuses
```
Mankro: The heading needs a space after the `#` characters. I'm not quite happy with the heading text. How about "List of exercise statuses".

Mankro: The list below should not be indented and there should be an empty line between the paragraph and the first list item. The explanations for the statuses could be expanded a little bit. E.g., "maintenance mode hides the exercise description and prevents submissions".

README.md
```diff
@@ -180,6 +180,7 @@ The questionnaire directive accepts the following options:
 * `pick_randomly`: integer. Set the pick_randomly setting for the quiz
   (select N questions randomly on each load)
 * `category`: exercise category
+* `status`: exercise status (default "unlisted"). See available [statuses](#available-statuses)
```
Mankro: Add a dot after the link (ending of the sentence).

README.md
```diff
@@ -239,7 +240,7 @@ after answering. See the example below.
     What is 1+1?
 
     a. 1
-    *b. 2
+    *b. 2[statuses](#available-statuses)
```
Mankro: This looks like an accident?

Mankro:
History has been squashed and I think this is good to go. @raphendyr do you still have comments about this? The PR does not validate the status values since validation ultimately belongs to another component.

Mankro:
@raphendyr last check that this is accepted for merging?

PR #38, opened 2019-07-25
The content of the submit directive to exercise instructions

alalav1:


alalav1:
The only relevant commit is the last one. I need to fix git history.

PR #39, opened 2019-07-25
The content of the submit directive to exercise instructions

alalav1:
The exercise instructions can now be written in the submit directives body. Instructions are written in rst and support most of it's features.

I had to redo this due to unclear git history.

### DIFF
--- a/directives/submit.py
+++ b/directives/submit.py
@@ -6,6 +6,7 @@ import os.path
 from docutils.parsers.rst import directives
 from docutils import nodes
 from sphinx.errors import SphinxError
+from sphinx.util.nodes import nested_parse_with_titles
 
 import aplus_nodes
 import lib.translations as translations
@@ -15,7 +16,7 @@ from directives.abstract_exercise import AbstractExercise
 
 
 class SubmitForm(AbstractExercise):
-    has_content = False
+    has_content = True
     option_spec = {
         'class' : directives.class_option,
         'quiz': directives.flag,
@@ -59,9 +60,6 @@ class SubmitForm(AbstractExercise):
         if 'ajax' in self.options:
             args[u'data-aplus-ajax'] = u'yes'
         node = aplus_nodes.html(u'div', args)
-        paragraph = aplus_nodes.html(u'p', {})
-        paragraph.append(nodes.Text(translations.get(env, 'submit_placeholder')))
-        node.append(paragraph)
 
         key_title = u"{} {}".format(translations.get(env, 'exercise'), key)
 
@@ -109,7 +107,23 @@ class SubmitForm(AbstractExercise):
             u'max_group_size': data.get('max_group_size', env.config.default_max_group_size),
             u'points_to_pass': self.options.get('points-to-pass', data.get('points_to_pass', 0)),
         })
+
+        if self.content:
+            self.assert_has_content()
+            nested_parse_with_titles(self.state, self.content, node)
+            # Sphinx can not compile the nested RST into HTML at this stage, hence
+            # the HTML instructions defined in this directive body are added to
+            # the exercise YAML file only at the end of the build. Sphinx calls
+            # the visit functions of the nodes in the last writing phase.
+            # The instructions are added to the YAML file in the depart_html
+            # function in aplus_nodes.py.
+        else:
+            paragraph = aplus_nodes.html(u'p', {})
+            paragraph.append(nodes.Text(translations.get(env, 'submit_placeholder')))
+            node.append(paragraph)
+
         data.setdefault('status', self.options.get('status', 'unlisted'))
+
         if category in override:
             data.update(override[category])
             if 'url' in data:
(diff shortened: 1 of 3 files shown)


directives/submit.py
```diff
@@ -4,18 +4,20 @@
 '''
 import os.path
 from docutils.parsers.rst import directives
+from docutils.parsers.rst import Directive
```
raphendyr:
```
from docutils.parsers.rst import (
    Directive,
    directives,
)
```

directives/submit.py
```diff
@@ -109,7 +111,18 @@ def run(self):
...
         })
+
+        if self.content:
+            self.assert_has_content()
+            instructions = '\n'.join(self.content)
+            try:
+                parts = docutils.core.publish_parts(source=instructions, writer_name='html')
+                data.update({u'instructions': parts['fragment']})
+            except Exception as e:
+                return str(e)
```
raphendyr: This seems relativle strange for exception handling. Where does the return go? Can Exception be limited to smaller scope than everything?

directives/submit.py
```diff
@@ -4,18 +4,20 @@
 '''
 import os.path
 from docutils.parsers.rst import directives
+from docutils.parsers.rst import Directive
 from docutils import nodes
 from sphinx.errors import SphinxError
 
+import docutils.core
```
raphendyr: maybe `from docutils.core import publish_parts`?

directives/submit.py
alalav1: I wasn't sure how to do this error handling. That return value doesn't go anywhere. Normally this submit directive doesn't do any kind of error handling. However, I am quite sure that errors of this rst->html conversion should be done here.

Mankro: Try this: `raise self.error('Error message.')`

alalav1: Like this?
`try:
  parts = ...
except Exception:
  raise self.error('Error message.')
else:
  data.update({...})`

Mankro: Yes.

directives/submit.py
```diff
@@ -109,6 +110,17 @@ def run(self):
...
             u'points_to_pass': self.options.get('points-to-pass', data.get('points_to_pass', 0)),
         })
+
+        if self.content:
+            self.assert_has_content()
+            instructions = '\n'.join(self.content)
+            try:
+                parts = publish_parts(source=instructions, writer_name='html')
+            except Exception:
+                raise self.error('Error in submit content!')
```
Mankro: The error message must contain info about the actual error instead of just saying there was some error. Something like 
```
except Exception as e:
    raise self.error(str(e))
```

Mankro:
These do not work in the submit body yet:

```
.. math::

  x = y^2 - z_2
```

The compilation gives a warning and the exercise is dropped out from the output.


```
Link to another page in the manual: 
:ref:`click <acos-docker-mac-windows>`
```

The compilation gives a warning and the exercise is dropped out from the output.

```
.. image:: /images/apluslogo.png
```

The image does not work because the image URL in the HTML output does not become a relative path inside the build directory as Sphinx normally does to images.


The API `docutils.core.publish_parts` may be unsuitable for parsing directive bodies. If it can't be made to work, then directive bodies are usually parsed with `sphinx.util.nodes.nested_parse_with_titles()`. Its problem is that it manipulates nodes instead of returning an HTML string. Thus, you need to extract the HTML output from the nodes somehow since we need the HTML for the YAML config file.

https://www.sphinx-doc.org/en/master/extdev/markupapi.html#parsing-directive-content-as-rest

raphendyr:
References inside exercise html should include full urls. Relatives can't work as the static html will be hosted on a different machine than the exercise itself.

Mankro:
References can work inside exercise HTML and they already do in questionnaires (written in RST). RST-tools manipulates the internal URLs so that A+ can fix the domain. The URLs then work in A+.

raphendyr:
For now, but that breaks when static material is hosted on different machine. Which is the plan, thus everything should start moving towards that.

Best url for A+ would be something like `https://{cdn-service}/course_id/course_code/full/path/to/fule.html` and then A+ would rewrite the `{}` part. Sadly, I don't think we will get there with current A+, so we need to build those urls in compile-rst step. CDN/static host url is provided in `STATIC_CONTENT_HOST` environment variable and is set in `conf.py` to `static_host = os.environ.get('STATIC_CONTENT_HOST')`.

Mankro:
I made a different solution without using directly the docutils API since it does not support all Sphinx features. Instead, it manipulates the YAML data of the nodes during the final writing phase of the Sphinx builder. Sphinx has normally produced the HTML of the submit directive body at that point and hence, Sphinx has resolved the internal links at that point too. The code can be viewed at the link for the moment.

https://github.com/Mankro/a-plus-rst-tools/commit/e03bec03e833ad6b56434250a9bc1514626f500c

Mankro:
We updated this PR to use my method instead of calling `docutils.core.publish_parts(source=instructions, writer_name='html')` since the docutils API does not support Sphinx-specific features.

Mankro:
The commits have been squashed and the README has been updated. I think this is ready for merging.

PR #40, opened 2019-08-08
Adding new options

alalav1:
Options allow-assistant-grading and allow-assistant-viewing can now be set in the submit and questionnaire directives.
Other commit is about clarifying the README about the usage of freetext questions in the feedback and normal questionnaires.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,10 @@ The questionnaire directive accepts the following options:
   (select N questions randomly on each load)
 * `category`: exercise category
 * `status`: exercise status (default "unlisted"). See available [statuses](#list-of-exercise-statuses).
+* `allow-assistant-viewing`: Allows assistants to view the submissions of the students.
+  Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
+* `allow-assistant-grading`: Allows assistants to grade the submissions of the students.
+  Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
 
 The contents of the questionnaire directive define the questions and possible
 instructions to students.
@@ -362,6 +366,10 @@ It accepts the following options:
   to the exercise and the exercise JS may control the submission itself.
   See [the chapter content documentation](https://github.com/Aalto-LeTech/a-plus/blob/master/doc/CONTENT.md)
   (the HTML attribute `data-aplus-ajax`).
+* `allow-assistant-viewing`: Allows assistants to view the submissions of the students.
+  Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
+* `allow-assistant-grading`: Allows assistants to grade the submissions of the students.
+  Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
 * `quiz`: If set, the exercise feedback will take the place of the exercise instructions.
   This makes sense for questionnaires since their feedback contains the submission form.
   In RST, you would usually define questionnaires with the questionnaire directive,
@@ -388,7 +396,7 @@ It accepts the following options:
 .. submit:: 2 A100
   :submissions: 100
   :config: exercises/hello_python/config.yaml
-  
+
   This will be shown in aplus as the instructions.
 ```
 
--- a/README.md
+++ b/README.md
@@ -227,7 +227,16 @@ For example, `.. freetext:: 30 string-ignorews-ignorequotes`.
 The question directives may define instructions. After the instructions,
 the contents of the directive define the choices, the correct solution, and
 possible hints. The hints are targeted to specific choices and they are shown
-after answering. See the example below.
+after answering.
+
+The body of the `freetext` question is
+expected to be its model solution. However, the question instructions can be written
+inside the body before the model answer. The instructions and the model solution must
+be separated with an empty line.
+
+If the questionnaire has the `feedback` option set, `freetext`
+questions may not have a model solution and the body of the question is shown as the
+question instructions.
 
 ```
 .. questionnaire:: 1 A60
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -181,6 +181,8 @@ The questionnaire directive accepts the following options:
   (select N questions randomly on each load)
 * `category`: exercise category
 * `status`: exercise status (default "unlisted"). See available [statuses](#list-of-exercise-statuses).
+* `allow-assistant-viewing`: Allows assistants to view the submissions of the students. Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
+* `allow-assistant-grading`: Allows assistants to grade the submissions of the students.Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
```
Mankro: One space is missing after one period. You don't need to write everything on a single line when the line becomes too long.

README.md
```diff
@@ -208,12 +210,19 @@ the common question options:
 The `freetext` directive accepts a second positional argument after the points.
 It defines the compare method for the model solution.
 A textual input can be compared with the model solution as `int`, `float`, `string`,
-or `unsortedchars` (unsorted character set).
+or `unsortedchars` (unsorted character set).
```
Mankro: New unnecessary space.

README.md
```diff
@@ -276,11 +285,14 @@ after answering. See the example below.
...
-A feedback questionnaire is basically just like a graded questionnaire, but with
-the `feedback` option it uses the feedback category and CSS class by default.
+A feedback questionnaire is almost like a graded questionnaire. With
+the `feedback` option questionnaire uses the feedback category and CSS class by default.
 The options `chapter-feedback`, `weekly-feedback`, `appendix-feedback`, and
 `course-feedback` use a different CSS class (with the same name as the option).
-If points are not specified, they are set to zero.
+If points are not specified, they are set to zero. The `feedback` option can be set only to the last questionnaire of the RST file.
+
+The freetext questions are not expected to have the model solution, in essence the
```
Mankro: One line is very long.

"the model solution" should be "a model solution" here.

README.md
```diff
@@ -276,11 +285,14 @@ after answering. See the example below.
 
 ### 2. Feedback questionnaire
 
-A feedback questionnaire is basically just like a graded questionnaire, but with
-the `feedback` option it uses the feedback category and CSS class by default.
+A feedback questionnaire is almost like a graded questionnaire. With
```
Mankro: With the `feedback` option questionnaire uses the feedback category and CSS class by default.
change that to
When the `feedback` option is set, the questionnaire uses the feedback category and CSS class by default.

README.md
```diff
@@ -362,6 +374,8 @@ It accepts the following options:
   to the exercise and the exercise JS may control the submission itself.
   See [the chapter content documentation](https://github.com/Aalto-LeTech/a-plus/blob/master/doc/CONTENT.md)
   (the HTML attribute `data-aplus-ajax`).
+* `allow-assistant-viewing`: Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
+* `allow-assistant-grading`: Can be set to true or false. Overrides any options set in the conf.py or config.yaml files.
```
Mankro: These lines seem to different here than in the other directive.

directives/questionnaire.py
```diff
@@ -120,6 +122,22 @@ def run(self):
                 raise SphinxError(u'Number of fields to sample randomly should greater than zero.')
             data[u'fieldgroups'][0]['pick_randomly'] = pick_randomly
 
+        if 'allow-assistant-grading' in self.options:
```
Mankro: Since the directive classes (Questionnaire and SubmitForm) inherit from AbstractExercise, you could move the identical code from the two classes into a new method in the super class. Then you just call the method when you parse the options `allow-assistant-viewing` and `allow-assistant-grading`.

README.md
```diff
@@ -214,6 +218,13 @@ expression and tries to match the submission with it.
 Strings have comparison modifiers that are separated with a hyphen.
 For example, `.. freetext:: 30 string-ignorews-ignorequotes`.
 
+The body of the `freetext` questionnaire is
```
Mankro: The paragraph about the body should be the last paragraph before the example code. Now it breaks the flow of the comparison modifiers since they are explained before this paragraph, but the list below is part of them.

Additionally, it is better to say "freetext questions" instead of "freetext questionnaires" since the questionnaire directive contains several questions.

directives/abstract_exercise.py
```diff
@@ -23,3 +23,21 @@ def extract_difficulty_and_points(self, arg):
                 else:
                     difficulty = u''.join(chars)
         return difficulty, points
+
+    def set_assistant_permissions(self,data):
```
Mankro: hmm there should a space `(self, data)`

README.md
```diff
@@ -276,11 +287,15 @@ after answering. See the example below.
...
 ### 2. Feedback questionnaire
 
-A feedback questionnaire is basically just like a graded questionnaire, but with
-the `feedback` option it uses the feedback category and CSS class by default.
+A feedback questionnaire is almost like a graded questionnaire. When the `feedback` option is set,
+the questionnaire uses the feedback category and CSS class by default.
 The options `chapter-feedback`, `weekly-feedback`, `appendix-feedback`, and
 `course-feedback` use a different CSS class (with the same name as the option).
-If points are not specified, they are set to zero.
+If points are not specified, they are set to zero.
```
Mankro: This one line seems to have one unnecessary space at the end.

Mankro:
There are still some small things to change and I also need to test the code on Monday.

directives/abstract_exercise.py
```diff
@@ -23,3 +23,21 @@ def extract_difficulty_and_points(self, arg):
                 else:
                     difficulty = u''.join(chars)
         return difficulty, points
+
+    def set_assistant_permissions(self, data):
+
+        if 'allow-assistant-grading' in self.options:
+            if str(self.options['allow-assistant-grading']).lower() == 'true':
+                data.update({'allow_assistant_grading': True})
```
Mankro: I forgot about the choice type in the option spec. Since this option has choices true or false, it is better to use the choice type instead of `directives.unchanged`. Then Sphinx/docutils will check if the user input matches any valid choice and you don't need this if statement.

directives/questionnaire.py
```diff
@@ -120,6 +122,8 @@ def run(self):
                 raise SphinxError(u'Number of fields to sample randomly should greater than zero.')
             data[u'fieldgroups'][0]['pick_randomly'] = pick_randomly
 
+        self.set_assistant_permissions(data)
```
Mankro: The modification to the data dict should be done before the `override` option (in conf.py) is handled because the `override` setting is supposed to be able to override other settings. Now the local `allow-assistant-viewing` value can not be overridden.

Mankro:
I made changes myself and pushed them so the issues have been fixed. @alalav1 you should rebase and squash the commits, but keep the assistant grading changes and the feedback changes in the README in separate commits.

Mankro:
and add me as the co-author in the commit message.

alalav1:
I have fixed the history.

PR #41, opened 2019-08-15
Fix questionnaires to use the title from the options if set there.

alalav1:
Previously questionnaires always used automatic titles.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -105,7 +105,6 @@ class Questionnaire(AbstractExercise):
             u'points_to_pass': self.options.get('points-to-pass', 0),
             u'feedback': is_feedback,
             u'view_type': u'access.types.stdsync.createForm',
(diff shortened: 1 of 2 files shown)


directives/questionnaire.py
```diff
@@ -124,6 +123,13 @@ def run(self):
                 raise SphinxError(u'Number of fields to sample randomly should greater than zero.')
             data[u'fieldgroups'][0]['pick_randomly'] = pick_randomly
 
+        if 'title' in self.options:
+            data['title'] = self.options.get('title')
+        else:
+            data[u'title|i18n'] = translations.opt('feedback') if is_feedback else translations.opt('exercise', postfix=u" {}".format(key))
+
+        self.set_assistant_permissions(data)
```
Mankro: I would prefer that the title is handled before the override if statement since the override setting is supposed to be able to override any configuration values. 

Additionally, the set assistant permissions method must not be duplicated here.

PR #43, opened 2019-08-26
Count maximum points for questionnaires.

alalav1:
Count the sum of the separate questions in questionnaire exercises. 

If the questionnaire has maximum points set, an error is raised if given points differ from the sum of individual questions. If the questionnaire doesn't have maximum points defined, the maximum points will be the sum of the points of the individual questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,10 +168,11 @@ html_theme_path = ['a-plus-rst-tools/theme']
 
 ### 1. Graded questionnaire
 
-The questionnaire directive arguments define the exercise key and max points
-with the optional difficulty. For example, `.. questionnaire:: 1 A50` sets key `1`,
-max points `50` and difficulty `A`.
-The questionnaire directive accepts the following options:
+The questionnaire directive arguments define the exercise key and optional max points
+with the difficulty. For example, `.. questionnaire:: 1 A50` sets key `1`,
+max points `50` and difficulty `A`. If not set in the directive arguments, the max points will be set to
+the sum of the question points. Setting the difficulty is optional and it can be set
+even if the max points aren't defined in the argument. The questionnaire directive accepts the following options:
 
 * `submissions`: max submissions
 * `points-to-pass`: points to pass
@@ -191,8 +192,8 @@ The contents of the questionnaire directive define the questions and possible
 instructions to students.
 
 The **question directives** `pick-one`, `pick-any`, and `freetext` take the points
-of the question as the first argument. The sum of the question points should
-be equal to the questionnaire max points.
+of the question as the first argument. If the questionnaire's max points are set, the sum
+of the question points should be equal to them.
 The question directives accept the following options:
 
 * `class`: CSS class
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -72,6 +72,7 @@ class Questionnaire(AbstractExercise):
 
         env.questionnaire_is_feedback = is_feedback
         env.question_count = 0
+        env.aplus_quiz_total_points = 0
 
         # Create document elements.
         node = aplus_nodes.html(u'div', {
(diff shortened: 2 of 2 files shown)


directives/questionnaire.py
```diff
@@ -149,6 +158,8 @@ class QuestionMixin:
     def create_question(self, title_text=None, points=True):
         env = self.state.document.settings.env
         env.question_count += 1
+        if points and len(self.arguments) > 0:
+            env.aplus_quiz_total_points += int(self.arguments[0])
```
Mankro: `env.aplus_quiz_total_points` could be updated later in the same if statement that saves the points in data (near the end of this method). It feels a little unnecessary to repeat the same if condition here.

directives/questionnaire.py
```diff
@@ -113,6 +114,14 @@ def run(self):
         }
         self.set_assistant_permissions(data)
 
+        points_set_in_arguments = False
+        if len(self.arguments) == 2 and difficulty != self.arguments[1]:
+            points_set_in_arguments = True
+
+        if points_set_in_arguments and env.aplus_quiz_total_points != points:
+                raise SphinxError('The points of the questions in the questionnaire must add up to the totalpoints of the questionnaire!')
```
Mankro: Spelling: totalpoints -> total points

directives/questionnaire.py
```diff
@@ -113,6 +114,14 @@ def run(self):
...
+        data['max_points']=env.aplus_quiz_total_points
```
Mankro: Add spaces around the `=` assignment operator.

Mankro: It looks like you could remove the `max_points` key from `data` when the dict is first initialized since the key is always set here again.

directives/questionnaire.py
Mankro: The `raise` line seems to have too much indentation.

README.md
```diff
@@ -168,9 +168,11 @@ html_theme_path = ['a-plus-rst-tools/theme']
...
 ### 1. Graded questionnaire
 
-The questionnaire directive arguments define the exercise key and max points
-with the optional difficulty. For example, `.. questionnaire:: 1 A50` sets key `1`,
-max points `50` and difficulty `A`.
+The questionnaire directive arguments define the exercise key and optional max points
+with the difficulty. For example, `.. questionnaire:: 1 A50` sets key `1`,
+max points `50` and difficulty `A`. If not set in the directive arguments, the max points will be set to
+the sum of the question points. Both difficulty and max points can be defined whether the other one is
+defined or not.
```
Mankro: "Both difficulty and max points can be defined whether the other one is defined or not."

I think this sentence is unclear and can be written in a better way. I assume you are saying that setting the difficulty is optional and the max points are taken from the sum of the question points if they are not manually defined in the argument.

directives/questionnaire.py
```diff
@@ -113,6 +113,14 @@ def run(self):
         }
         self.set_assistant_permissions(data)
 
+        points_set_in_arguments = False
+        if len(self.arguments) == 2 and difficulty != self.arguments[1]:
+            points_set_in_arguments = True
+
+        if points_set_in_arguments and env.aplus_quiz_total_points != points:
+            raise SphinxError('The points of the questions in the questionnaire must add up to the total points of the questionnaire!')
```
Mankro: The error message must show to the user where the problem is, i.e., the RST file and line number of the problematic questionnaire. At least this method is one way to obtain that information: `self.state_machine.get_source_and_line(self.lineno)`. You need to format the error message so that it shows the location in the beginning.

alalav1: something like this?
```
            source, line = self.state_machine.get_source_and_line(self.lineno)
            raise SphinxError("Error in " + source +" in line "+ str(line)+ ":\nThe points of the questions in the questionnaire must add up to the totalpoints of the questionnaire!")
```

Mankro:
@alalav1 please squash the commits.

PR #44, opened 2019-08-27
Add two options to the POI directive for the presentation maker

zunde:
These options are needed for the new presentation maker tool that generates lecture slides from the RST materials. The slide contents are taken from the point of interest directives.

### DIFF
--- a/directives/point_of_interest.py
+++ b/directives/point_of_interest.py
@@ -11,6 +11,8 @@ Directive for creating "point of interest" summary block.
     :height: optional fixed height for content
     :bgimage: path to background image
     :columns: relative widths of poi content columns (e.e. 2 3 3)
+    :not_in_slides: used with the presentation maker. This POI does not show in the slides when used.
+    :not_in_book: This POI does not appear in the book material when used
 
     Content of point-of-interest here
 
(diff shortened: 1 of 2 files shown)


directives/point_of_interest.py
```diff
@@ -115,97 +120,102 @@ def run(self):
...
-        poi_info = {
-            'docname': env.docname,
-        }
-        if 'previous' in self.options:
-            poi_info['previous'] = self.options['previous']
-        if 'next' in self.options:
-            poi_info['next'] = self.options['next']
-
-        env.poi_all[name] = poi_info
+        if not 'not_in_book' in self.options:
```
Mankro: Instead of putting so much code into a simple if statement, it is better to check the condition and return early from the run method. Then, there is no need to indent so many existing lines.

For example, I think you could do this right after the `node` variable has been defined:

```
if 'not_in_book' in self.options:
    poi_info = {
        'docname': env.docname,
    }
    env.poi_all[name] = poi_info
    return [node]
```

Mankro:
@zunde Please simplify the code according to the comment above and then squash your commits into a single commit.

PR #46, opened 2019-09-10
Add settings to be defined in the index.rst or conf.py

alalav1:
These course-wide settings can be defined in main index.rst inside an aplusmeta directive. The values written there will be prioritized over those written in the conf.py. All settings doesn't need to be defined in the same place.

### DIFF
--- a/aplus_setup.py
+++ b/aplus_setup.py
@@ -43,7 +43,19 @@ def setup(app):
     app.add_config_value('course_head_urls', None, 'html')
     app.add_config_value('bootstrap_styled_topic_classes', 'dl-horizontal topic', 'html')
     app.add_config_value('acos_submit_base_url', 'http://172.21.0.2:3000', 'html')
-
+    app.add_config_value('numerate_ignoring_modules', None, 'html')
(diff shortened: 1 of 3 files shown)


Mankro:
One idea was that the key-value pairs of RST documents (known as "field lists" in the RST specification) could be used to define these course settings in the main index.rst. However, it is hard to find out how those field lists could be used in the Python code in a-plus-rst-tools. Basically, we can not find the API for them. Additionally, docutils/Sphinx seems to include the field lists in the HTML output too, which is not really what we want since the course configurations are not part of the course contents (though, we do not use the content of the main index.html anywhere so then it matters less).

Link to the documentation of field lists:
http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#field-lists

Mankro:
PR https://github.com/Aalto-LeTech/a-plus-rst-tools/pull/48 was merged and chosen over this PR so this one is closed now. 

The older related PR https://github.com/Aalto-LeTech/a-plus-rst-tools/pull/45 is also closed.

PR #47, opened 2019-09-12
Modify the grading of pick-any questions

alalav1:
Add the option `partial-points` and add neutral choice. Most of the changes are made
in the mooc-grader in commit https://github.com/Aalto-LeTech/mooc-grader/pull/42

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,10 +274,19 @@ possible hints. The hints are targeted to specific choices and they are shown
 after answering.
 
 Correct answers in `pick-one` and `pick-any` directives are marked with `*`.
-Initially selected values may be set with `+`. If an option is both initially
-set and correct, `+` is written before `*`. The `pick-one` questions are
-rendered with HTML radio buttons by default, but a dropdown (select) element
-may be used with the `dropdown` option.
+A `pick-any` question may have neutral options, which are marked with `?`.
+Neutral options are always counted as correct, whether the student selected them
+or not. Initially selected options may be set with `+`. The initially selected
+options are pre-selected when the exercise is loaded. The `+` character is
+written before `*` or `?` if they are combined.
+
+The `pick-any` directive has a `partial-points` option in order to award points
+for partially correct submissions. Zero points are awarded to answers that mark
+half or less of the options (checkboxes) correctly. The points scale linearly to
+the maximum points when more than half of the options are answered correctly.
+
+The `pick-one` questions are rendered with HTML radio buttons by default, but
+a dropdown (select) element may be used with the `dropdown` option.
 
 The body of the `freetext` question is
 expected to be its model solution. However, the question instructions can be written
@@ -293,7 +302,7 @@ question instructions.
   :submissions: 4
   :points-to-pass: 0
 
-  This is a questionnaire number 1 that grants at maximum 70 points
+  This is a questionnaire with the key `1` that grants at maximum 70 points
   of difficulty A. Students can make at most 4 submissions.
   This exercise is marked passed when 0 points are reached (the default).
 
@@ -307,6 +316,7 @@ question instructions.
     c. 3
 
     !b § Count again!
+    b § That is correct!
     c § Too much
 
   (Hints can be included or omitted in any question.)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,6 +202,11 @@ The question directives accept the following options:
   the key in the submission data. If no key is set, note that automatically added
   keys change when the order and amount of questions is modified.
 
+The `pick-any` directive has a `partial-points` option. When `partial-points` is
+set, some of the points are granted if the submission is partially correct.
```
Mankro: You should briefly explain how partial points are calculated (at least half of the checkboxes must be correctly selected etc.).

directives/questionnaire.py
```diff
@@ -313,6 +315,9 @@ def split_second_last(empty_lines):
             if key.startswith(u'*'):
                 correct = True
                 key = key[1:]
+            if key.startswith(u'?'):
```
Mankro: I suppose the `key` should never have both `*` and `?`, so if statement could use `elif`.

directives/questionnaire.py
```diff
@@ -387,7 +395,6 @@ class FreeText(QuestionMixin, Directive):
         'no-standard-prompt': directives.flag,
         'shorter-prompt': directives.flag,
         'class': directives.class_option,
-        'required': directives.flag,
```
Mankro: I don't think you can remove the `required` option here. This dict does not inherit anything from the super class QuestionMixin since the value is overridden here. It would make sense to inherit the common options, but it would require code like this:

```
option_spec = dict(QuestionMixin.option_spec)
option_spec.update({ .... }) # new key-value pairs
```

directives/questionnaire.py
```diff
@@ -361,6 +366,9 @@ def input_type(self):
 
 class MultipleChoice(Choice):
     ''' Lists options for picking all the correct ones. '''
+    # inherit QuestionMixin options and add a key
+    option_spec = dict(QuestionMixin.option_spec)
+    option_spec.update({'partial-points': directives.unchanged})
```
Mankro: Ǹo need to use the `update` method for a single key. Do it like this: 
`option_spec['partial-points'] = directives.flag`

This is supposed to be a flag, right? So use `directives.flag`.

Mankro:
@alalav1 TODO list
1. rebase against the upstream master branch, 
2. fix the git conflicts, 
3. fix the small issues that I commented about and 
4. squash the commits into a single commit.

directives/questionnaire.py
alalav1: `required` option has been written twice in that option dict, so there still would be one of them.

Mankro:
@alalav1 The commit message should be written better and more clearly. Describe the new features briefly there.

README.md
```diff
@@ -244,6 +244,12 @@ The question directives accept the following options:
   the key in the submission data. If no key is set, note that automatically added
   keys change when the order and amount of questions is modified.
 
+The `pick-any` directive has a `partial-points` option to award points for partially correct
+submissions. The zero point threshold is when half of the checkboxes are marked correctly, and
+additional correct marks raise the points linearly to the max points.
+A `pick-any` question may have neutral fields, which can be marked with `?`.
```
Mankro: The neutral fields with `?` should be described below in the same paragraph as the notation for correct answers with `*`. The partial points option can be moved there as well.

README.md
```diff
@@ -329,6 +335,7 @@ question instructions.
     *a. this is the **first**
     *b. this is the **second**
     c. this is the **third**
+    ?d. choosing this will not affect to the granted points
```
Mankro: `choosing this will not affect to the granted points` should be `choosing this does not affect the granted points`

Mankro: There should also be an example of a partial points question. When you add a new question, update the max points in the beginning of the questionnaire too.

directives/questionnaire.py
```diff
@@ -353,6 +355,9 @@ def split_second_last(empty_lines):
             if key.startswith(u'*'):
                 correct = True
                 key = key[1:]
+            elif key.startswith(u'?'):
```
Mankro: No need to add `u` to the string.

directives/questionnaire.py
```diff
@@ -361,7 +366,7 @@ def split_second_last(empty_lines):
                 'value': key,
             }
             if correct:
-                optdata['correct'] = True
+                optdata[u'correct'] = correct
```
Mankro: No need to add `u` to the string.

directives/questionnaire.py
```diff
@@ -433,6 +438,9 @@ def input_type(self):
 
 class MultipleChoice(Choice):
     ''' Lists options for picking all the correct ones. '''
+    # inherit QuestionMixin options and add a key
```
Mankro: Add an empty line after the docstring `''' Lists options .... '''` and start your new comment with the capital letter `I` and end the sentence with the period `.`.

Mankro:
The PR was closed by accident and then re-opened. I made some changes to the README and force-pushed before merging.

PR #48, opened 2019-09-13
Course settings can be set in main doctree as field-list

alalav1:


alalav1:
This PR is parallel to #46 , but in this version values can't be set in conf.py and course settings are written as field-list in the main toc-tree.
```
:course-start: 2019-10-10
```

PR #50, opened 2019-09-18
Add  'dropdown'  option to pick-one questions

alalav1:
Add 'dropdown' option to pick-one questions.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -2,6 +2,8 @@
 '''
 Directives that define automatically assessed questionnaires.
 '''
+import html
+
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
 from sphinx.errors import SphinxError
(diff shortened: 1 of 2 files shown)


directives/questionnaire.py
```diff
@@ -349,13 +349,21 @@ def split_second_last(empty_lines):
 class SingleChoice(Choice):
     ''' Lists options for picking the correct one. '''
 
+    # inherit option_spec from QuestionMixin and add a key
+    option_spec = dict(QuestionMixin.option_spec)
+    option_spec.update({'dropdown': directives.flag})
```
Mankro: Could it be possible to define `option_spec` in a single statement?

Mankro:
You could check if it is possible to output multiselect HTML widget for a `pick-any` question with the dropdown option.

Mankro:
The pick-any multiselect HTML widget did not seem to feasible now, so ignore that.

Mankro:
README needs to be updated too.

Mankro:
The dropdown (select) element did not work correctly in the mock frame (built html files in the `_build` directory), but I fixed it myself. The HTML exercise descriptions built into the mock frame are not used by the MOOC grader since the MOOC grader creates the HTML itself from the config YAML files, but some people like to preview the exercise description in the mock frame. 

Additionally, I made other small stylistic changes. I force pushed after rebasing and squashing so that I could immediately merge this PR.

PR #49, opened 2019-09-16
Bugfix: language is now read from the language postfix in multilingual courses

alalav1:
Previously all index.yaml files used the same language, due to reading it from conf.py (even if all languages were set there, there was no method to choose right language for index_lang.yaml).
Trello issue: https://trello.com/c/F3iqx2wp/25-aplus-rst-tools-should-set-appenvconfiglanguage-appropriately-during-setup-so-that-translationspy-actually-does-its-job

PR #52, opened 2019-10-01
Fix: in multilingual courses None titles in yaml files removed

alalav1:
Previously if `title|i18n` was determined in config.yaml, build files often
contained None as title value as well. Now the key `title` is not written in the yaml if not defined in the RST file. Additionally if the `title|i18n` exists the handwritten titles in the RST are saved
into the `title|i18n` itself, so there can never be both `title|i18n` and `title` written in a yaml file.
If `env.conf.language` is not correct,
the manually written title is saved in under the wrong language, yet in the correct
yaml, so the written title doesn't override anything.

PR about defining `env.conf.language` #49

### DIFF
--- a/directives/submit.py
+++ b/directives/submit.py
@@ -71,7 +71,7 @@ class SubmitForm(AbstractExercise):
             if not os.path.exists(path):
                 raise SphinxError('Missing config path {}'.format(self.options['config']))
             data = yaml_writer.read(path)
-            config_title = data.get(u'title', None)
+            config_title = data.get('title', '')
         else:
             data = { u'_external': True }
             if 'url' in self.options:
@@ -86,7 +86,7 @@ class SubmitForm(AbstractExercise):
                     data.update({u'lti_aplus_get_and_post': True})
                 if 'lti_open_in_iframe' in self.options:
                     data.update({u'lti_open_in_iframe': True})
-            config_title = None
+            config_title = ''
 
         config_title = self.options.get('title', config_title)
         if "radar_tokenizer" in self.options or "radar_minimum_match_tokens" in self.options:
@@ -98,9 +98,6 @@ class SubmitForm(AbstractExercise):
         category = u'submit'
         data.update({
             u'key': name,
-            u'title': env.config.submit_title.format(
-                key_title=key_title, config_title=config_title
-            ),
             u'category': u'submit',
             u'scale_points': points,
             u'difficulty': difficulty or '',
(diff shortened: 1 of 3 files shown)


Mankro:
This is interesting if this small change could fix the issue with titles. However, there may be other `i18n` fields that could suffer from the same bug. I was actually thinking that the `lib/toc_languages.py` file could need changes due to this issue. In addition, the title issue may well affect questionnaires too (I think we already saw that in one course) and then changing the submit directive is not enough.

alalav1:
In the situation where the handwritten exercise yaml has both `title|i18n` and `title` defined, the build yaml will also have them both. This shouldn't be usual, and warning could be given of such usage, since it doesn't make sense to use both type of titles.

Mankro:
We have one multilingual course that defines exercises in RST so that both language versions link to the same config.yaml file. That is why the config.yaml defines `title|i18n` for both languages in the same YAML file. The problem is that the compiled YAML file also has `title: None`. A-plus-rst-tools should not write the `title: None` field to the built output YAML file when the course has defined `title|i18n` in its exercise config.yaml.

alalav1:
> This is interesting if this small change could fix the issue with titles. However, there may be other `i18n` fields that could suffer from the same bug. I was actually thinking that the `lib/toc_languages.py` file could need changes due to this issue. In addition, the title issue may well affect questionnaires too (I think we already saw that in one course) and then changing the submit directive is not enough.

Questionnaire titles in multilingual courses were modified in #41 . The problem was that titles were not read from the options, even if set. I tested it on multilingual courses. The language-specific yamls have `title` if option set in the RST file or automatic title in `title|i18n` . I just tested that the yaml files work correctly, so that `title` is never set as None in main yaml.

Mankro:
Title: None actually implies that Python has written the string "None", since Python None becomes `null` in YAML. The submit and questionnaire directives have some weird special cases that can write title "None" like @alalav1 has said, so it is necessary to change the directive code to handle titles better. Like @alalav1 has started. :+1:

directives/submit.py
```diff
@@ -111,6 +108,14 @@ def run(self):
         })
         self.set_assistant_permissions(data)
 
+        if data.get('title|i18n'):
+            if config_title:
+                data['title|i18n'][env.config.language] = config_title
+        else:
+            data['title'] = env.config.submit_title.format(
+                key_title=key_title, config_title=config_title
+            )
```
alalav1: Currently the `env.config.submit_title` uses always the `config_title` in the default version of a-plus-rst-tools. When the `config_title` is not defined, string 'None' is used. I think this is never the way the `title` is initiated to behave. By default rst-tools could use `key_title`, if `config_title` is not defined.

Mankro:
I made some changes, force pushed and merged.

Mankro:
@raphendyr FYI, this PR fixes the `title: None` issue that affected courses that had defined `title|i18n` in their exercise config.yaml files. The submit directive now simply does not write `title` to the built yaml file if the config.yaml already contains `title|i18n`.

PR #53, opened 2019-10-07
Fix points of pick_randomly questionnaires

alalav1:
Require question points to be equal when `pick_randomly` is set. Count maximum
points into the `points_calculated` variable.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -7,12 +7,16 @@ import html
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
 from sphinx.errors import SphinxError
+from sphinx.util import logging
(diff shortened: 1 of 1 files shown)


directives/questionnaire.py
```diff
@@ -74,7 +74,9 @@ def run(self):
 
         env.questionnaire_is_feedback = is_feedback
         env.question_count = 0
+        env.single_question_points = None
```
Mankro: The `env` variables should be prefixed with the extension name `aplus_` since `env` is shared between all extensions. We have to minimize the risk that other extensions use the same names. So, `env.single_question_points` -> `env.aplus_single_question_points`.

Mankro:
I made some stylistic changes, force-pushed and merged.

PR #54, opened 2019-10-14
No need to set separate questionpoints in a feedback questionnaire.

alalav1:
If none of the questions in a feedback questionnaire has points set, the questionnaire can still have non-zero maxpoints.


Mankro:
My force-push only changed the commit message.

PR #56, opened 2019-10-18
Added functionality for "annotated" directive

jsorva:
This PR adds two things: 

1) inlined annotations (rendered as spans not divs); 

2) replacement texts that can be added to annotations as html attributes (and can be rendered by front-end code however it pleases)

In practice, this merely adds new possibilities for course authors and should not impact any existing courses in production.

### DIFF
--- a/directives/annotated.py
+++ b/directives/annotated.py
@@ -33,6 +33,8 @@ def slicer(stringList):
 
 class annotated_node(nodes.General, nodes.Element): pass
 
+inline_anno_pattern = r"\[\[\[([^¶]+?)(?:¶(.*?))?\]\]\]"   # [[[annotation]]] or [[[annotation¶replacement]]]
+
 class AnnotatedSection(Directive):
     has_content = True
     required_arguments = 0
@@ -56,6 +58,10 @@ class AnnotatedSection(Directive):
         highest_annotation = self.assert_sanity(self.block_text)
         if not highest_annotation:
             return [self.state.document.reporter.error('Invalid annotation markers embedded in ' + self.block_text)]
+
+        # Inline annotations numbered first (before nested_parse deals with annotation directives)
+        inline_anno_count = len(re.findall(inline_anno_pattern, self.block_text))
+        env.annotated_annotation_count += inline_anno_count
 
         self.state.nested_parse(self.content, 0, node)
         node['name'] = env.annotated_name
@@ -87,11 +93,27 @@ def depart_annotated_node(self, node):
     parsed_html = self.body  # extract generated feedback line
     self.body = env.redirect # restore original output
 
-    self.body.append(postprocess(u''.join(parsed_html), node['name']))
+    postprocessed_html = postprocess_annotation_tags(u''.join(parsed_html), node['name'])
+    postprocessed_html = postprocess_inline_annotations(postprocessed_html, node['name'])
+    self.body.append(postprocessed_html)
 
     self.body.append("</div>\n")
 
-def postprocess(html, annotation_id):
+def postprocess_inline_annotations(html, annotated_section_id):
+    inline_anno_count = 0
+
+    def make_annotation_span(match):
+        nonlocal inline_anno_count
+        inline_anno_count += 1
+        annotation_text = match.group(1)
+        bit_to_insert = match.group(2)
+        replacement_attrib = ' data-replacement="' + bit_to_insert + '"' if bit_to_insert else ""
+        html_bits = (annotated_section_id, inline_anno_count, replacement_attrib, annotation_text)
+        return '<span class="codecomment comment-%s-%s"%s>%s</span>' % html_bits
+
+    return re.sub(inline_anno_pattern, make_annotation_span, html)
+
+def postprocess_annotation_tags(html, annotation_id):
     processed   = []
     openstack   = []
     selfclosing = []
--- a/directives/annotated.py
+++ b/directives/annotated.py
@@ -2,7 +2,7 @@
 import docutils
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
-from cgi import escape
+from html import escape
 from collections import Counter
 import re
 import os
@@ -137,7 +137,8 @@ class annotation_node(nodes.General, nodes.Element): pass
 class Annotation(Directive):
     has_content = True
     required_arguments = 0
-    optional_arguments = 0
+    optional_arguments = 1
+    final_argument_whitespace = True
     option_spec = { }
 
     def run(self):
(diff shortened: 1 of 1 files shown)


directives/annotated.py
```diff
@@ -87,11 +92,26 @@ def depart_annotated_node(self, node):
     parsed_html = self.body  # extract generated feedback line
     self.body = env.redirect # restore original output
 
-    self.body.append(postprocess(u''.join(parsed_html), node['name']))
+    postprocessed_html = postprocess_annotation_tags(u''.join(parsed_html), node['name'])
```
Mankro: Unicode literals are not needed anymore since we only use Python 3: `u''` -> `''` (remove u)

directives/annotated.py
```diff
@@ -87,11 +92,26 @@ def depart_annotated_node(self, node):
...
-def postprocess(html, annotation_id):
+def postprocess_inline_annotations(html, annotated_section_id):
+    inline_anno_count = 0
+
+    def make_annotation_span(match):
+        nonlocal inline_anno_count
+        inline_anno_count += 1
+        annotation_text = match.group(1)
+        bit_to_insert = match.group(2)
+        insert_attrib = ' insert="' + bit_to_insert + '"' if bit_to_insert else ""
```
Mankro: Seems like there is an HTML attribute `insert` here. It is not a standard attribute, so it should be `data-insert`. Data attributes are intended for custom data storage.

directives/annotated.py
```diff
@@ -153,10 +174,13 @@ def run(self):
         env.annotated_annotation_count += 1
         node['annotation-number'] = env.annotated_annotation_count
         node['name-of-annotated-section'] = env.annotated_name
+        if self.arguments:
+            node['replacement'] = self.arguments[0]
         return [node]
 
 def visit_annotation_node(self, node):
-    self.body.append('<div class="container codecomment comment-%s-%s">' % (node['name-of-annotated-section'], node['annotation-number']))
+    replacement_attrib = ' replacement="' + node['replacement'] + '"' if 'replacement' in node else ""
```
Mankro: Seems like there is an HTML attribute `replacement` here. It is not a standard attribute, so it should be `data-replacement`. Data attributes are intended for custom data storage. (This is only about the HTML: Sphinx nodes may use whatever attributes.)

directives/annotated.py
Mankro: That was only about the HTML. Sphinx/docutils nodes can use any arbitrary attributes.

directives/annotated.py
jsorva: OK. I didn't actually edit that call in this PR, but I fixed that line now. At some point, we could remove all the unnecessary u's from the repo.

directives/annotated.py
jsorva: Thanks. I committed a change that uses a better name for that attribute throughout.

directives/annotated.py
```diff
@@ -87,11 +92,27 @@ def depart_annotated_node(self, node):
...
+    inline_anno_count = 0
+
+    def make_annotation_span(match):
+        nonlocal inline_anno_count
+        inline_anno_count += 1
+        annotation_text = match.group(1)
+        replacement = match.group(2)
+        replacement_attrib = ' data-replacement="' + replacement + '"' if replacement else ""
+        html_bits = (annotated_section_id, inline_anno_count, replacement_attrib, annotation_text)
+        return '<span class="codecomment comment-%s-%s"%s>%s</span>' % html_bits
```
Mankro: I have a feeling that the `annotation_text` and `replacement` values should be HTML escaped before writing them to HTML. If the RST content contains text like `5 > 3`, the `>` character could cause issues in the HTML if it is not escaped.

Mankro: https://docs.python.org/3/library/html.html#html.escape

directives/annotated.py
```diff
@@ -153,10 +175,14 @@ def run(self):
...
         node['name-of-annotated-section'] = env.annotated_name
+        if self.arguments:
+            node['replacement'] = self.arguments[0]
         return [node]
 
 def visit_annotation_node(self, node):
-    self.body.append('<div class="container codecomment comment-%s-%s">' % (node['name-of-annotated-section'], node['annotation-number']))
+    replacement_attrib = ' data-replacement="' + node['replacement'] + '"' if 'replacement' in node else ""
+    html_bits = (node['name-of-annotated-section'], node['annotation-number'], replacement_attrib)
+    self.body.append('<div class="container codecomment comment-%s-%s"%s>' % html_bits)
```
Mankro: The same thing about HTML escaping as above: https://docs.python.org/3/library/html.html#html.escape

Mankro:
Before merging the PR, it is nice to squash the small fixes into the "main" commits (the two original commits).

directives/annotated.py
jsorva: Unless I'm mistaken, that isn't an issue, because this code runs only after the contents of the `annotated::` directive have been processed into HTML (and escaped). At that point, `annotation_text` and `replacement` are part of that HTML output and this bit just relocates them within that putput. 

"Works fine when I tried it." ;-)

directives/annotated.py
jsorva: Hmm, unlike the above (which is an inlined annotation), this is a bona-fide RST directive, so its argument doesn't get parsed. So, yeah, escaping could be an issue if someone wants to use double-quotes here.

Mankro: Indeed. The annotation argument must be HTML-escaped. Double quotes absolutely break the HTML attribute. Other special HTML characters must be escaped too since there are cases in which they mess up HTML parsers inside attributes. Calling `html.escape()` is enough.

directives/annotated.py
Mankro: I agree. Sphinx escapes the HTML here since the function processes the output of other nodes that are nested inside the annotated node. So no escaping should be added here.

Mankro:
It looks like that the inline annotations (`[[[  ]]]`) are always numbered first before the `.. annotation::` directives. This should be clarified somewhere since it affects the numbering inside the annotated section (the `1«code»` parts).

Mankro:
I found an escaping issue, but I decided to ignore it now and merge this anyway.

If the annotated section contains the following inline annotation, the manually typed `<script>` is HTML-escaped, but the HTML `<strong>` compiled from the RST `**bold**` (in the replacement attribute) is of course not escaped. Since this value is output into an HTML attribute, it should be escaped. That is to say, the replacement section in `[[[  ]]]` after the `¶` symbol should not use any RST formatting or else there could be issues with the missing HTML escaping.

RST
```
TESTING inline annotation: [[[text here with **some bold**]]] and then try the replacement
[[[using *italics* and something to HTML escape 2 > 1¶TEST REPLACE content **bold** and escape <script>]]]
```

HTML with the attribute that contains `<strong>` that should have been escaped with `&lt;` etc.
```
<p>TESTING inline annotation: <span class="codecomment comment-w02ch01_fi_1-1">text here with <strong>some bold</strong></span> and then try the replacement
<span class="codecomment comment-w02ch01_fi_1-2" data-replacement="TEST REPLACE content <strong>bold</strong> and escape &lt;script&gt;">using <em>italics</em> and something to HTML escape 2 &gt; 1&nbsp;</span></p>

```

jsorva:
OK, thanks for merging and pointing that out.

PR #57, opened 2019-11-11
Add an option to show the correct answer after the last submission

alalav1:
Add `show_correct_answer` -option to questionnaires to show model solution. Previously the correct
answer was always shown after the last submission, but PR to change that is already in mooc-grader Aalto-LeTech/mooc-grader#48.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -13,7 +13,7 @@ from sphinx.util.nodes import nested_parse_with_titles
 import aplus_nodes
 import lib.translations as translations
 import lib.yaml_writer as yaml_writer
-from directives.abstract_exercise import AbstractExercise, choice_truefalse
+from directives.abstract_exercise import AbstractExercise, choice_truefalse, str_to_bool
 
 
 logger = logging.getLogger(__name__)
@@ -35,6 +35,7 @@ class Questionnaire(AbstractExercise):
         'title': directives.unchanged,
         'category': directives.unchanged,
         'status': directives.unchanged,
+        'reveal-model-at-max-submissions': choice_truefalse,
         'allow-assistant-viewing': choice_truefalse,
         'allow-assistant-grading': choice_truefalse,
     }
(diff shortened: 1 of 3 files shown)


Mankro:
You don't really mean that the correct model answer was **always** shown, right? It was only shown in the feedback after the student had consumed the available submissions.

aplus_setup.py
```diff
@@ -44,6 +44,7 @@ def setup(app):
     app.add_config_value('course_head_urls', None, 'html')
     app.add_config_value('bootstrap_styled_topic_classes', 'dl-horizontal topic', 'html')
     app.add_config_value('acos_submit_base_url', 'http://172.21.0.2:3000', 'html')
+    app.add_config_value('show_correct_once', False, 'html')
```
Mankro: Is the config value really needed at all? If the MOOC-grader does not show the model solution by default, then the rst-tools does not need to write any default value for the config key.

directives/questionnaire.py
```diff
@@ -35,6 +35,7 @@ class Questionnaire(AbstractExercise):
         'title': directives.unchanged,
         'category': directives.unchanged,
         'status': directives.unchanged,
+        'show-correct-once': choice_truefalse,
```
Mankro: Would `directives.flag` work here if the default is False and the True value is only written to the config YAML if the directive has the option flag set?

Mankro:
`show-correct-once` is a confusing and misleading name for the option. What would be more descriptive when this should control whether students see the model answer after submitting the maximum number of times?

You used a merge commit and that is why GitHub can not rebase these commits onto master. You should rebase your branch on the upstream master branch.

directives/questionnaire.py
alalav1: But if the teacher wants to set default to the course that model answers are shown but wants to prevent that for some specific exercises, there should be an easy way to do that, too.

Mankro: OK. Let's move the default setting from conf.py to the field list of the index.rst, though.

Mankro:
I changed the name of the options and the implementation of the string conversion to a boolean.

PR #58, opened 2019-11-15
Fix collisions of the chapter keys in nested directories.

alalav1:
This commit fixes the chapter keys in nested directories, so that they will
always be unique. The links to nested chapters need to be fixed as well. Update: This PR is about fixing those links, too.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -122,6 +122,9 @@ class Questionnaire(AbstractExercise):
                 u'title': '',
                 u'fields': (u'#!children', None),
             }],
+            # The RST source file path is needed for fixing relative URLs
+            # in the exercise description.
+            '_rst_srcpath': env.doc2path(env.docname, None),
         }
 
         meta_data = env.metadata[env.app.config.master_doc]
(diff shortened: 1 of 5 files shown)


alalav1:
So far the nested chapter keys have been fixed, the links to nested chapter in monolingual courses work as they should. The links in multilingual courses work when: 
- Chapter is not in the nested directory
- The chapter keys are form `name+'_en'` where the name is same for every language and does not contain characters `_` and `-`.
TODO: Fix nested directory links in multilingual courses and allow using of `-` and `-`.

alalav1:
> So far the nested chapter keys have been fixed, the links to nested chapter in monolingual courses work as they should. The links in multilingual courses work when:
> 
>     * Chapter is not in the nested directory
> 
>     * The chapter keys are form `name+'_en'` where the name is same for every language and does not contain characters `_` and `-`.
>       TODO: Fix nested directory links in multilingual courses and allow using of `_` and `-` in chapter names.

Also investigate if links in multilingual courses could work for all chapter names.

alalav1:
It seems that chapters in nested directories actually work as they are supposed to. Furthermore the earlier mention of the key modification removing separators `-` and `_` was not quite accurate. The key modification only removes consecutive separators so keys work as intended if the key doesn't have consecutive separators. There is still one more condition in the naming: chapter names shouldn't start with undescore '_' and nested folders shouldn't end with underscore '_' in multilingual courses. This doesn't crash the site, but links pointing to these chapters don't work.

Mankro:
Since chapter keys are built differently than before, we need to check if there are issues with them in old courses already in production. If the new keys are the same as before in the normal case, then there shouldn't be issues (normal case = no nested directories under the module directory).

PR #59, opened 2019-11-22
Improved questionnaire feedback and hints

alalav1:
Submitter may receive hints based on matching regular expression. Regexp hints work similarly to the previous hints and are started with `regexp:`. README will have an example soon enough.

Add `subdiff` compare method to freetext questions. It works almost like normal `string` compare method with some features:
- It may have multiple correct answers, delimited with `|`.
- Always shows the difference to each correct answer.

For example if correct answer is 'superman' and student answers 'steelman', the feedback contains string: `Correct parts in your answer: s----man`

Most of the modifications have been made in mooc-grader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,23 +267,34 @@ the common question options:
 
 The `freetext` directive accepts a second positional argument after the points.
 It defines the compare method for the model solution.
-A textual input can be compared with the model solution as `int`, `float`, `string`,
-or `unsortedchars` (unsorted character set).
-Another option is `regexp` that takes the correct answer as a regular
+A textual input can be compared with the model solution as `int`, `float`,
+`string`, `subdiff`, `regexp` or `unsortedchars` (unsorted character set).
+The `regexp` compare method takes the correct answer as a regular
 expression and tries to match the submission with it.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -264,10 +264,13 @@ the common question options:
...
-A textual input can be compared with the model solution as `int`, `float`, `string`,
-or `unsortedchars` (unsorted character set).
-Another option is `regexp` that takes the correct answer as a regular
+A textual input can be compared with the model solution as `int`, `float`,
+`string`, `subdiff`, `regexp` or `unsortedchars` (unsorted character set).
+The `regexp` compare method takes the correct answer as a regular
 expression and tries to match the submission with it.
+The `subdiff` method works almost like the `string` method, but it can have
+multiple correct answers separated with `|` and it shows difference to the each
+correct answer using pythons difflib.
```
Mankro: Instead of referring to the "pythons difflib" that the reader probably knows nothing about, we should explain what kind of feedback is given. "If the submission is incorrect, the feedback shows its difference to the correct solutions. For example, ..."

PR #61, opened 2019-12-04
Add postfixes to links so that sphinx finds the files

alalav1:
Courses with language postfixes in their file names such as `first_round_en`
and `first_round_fi` can now leave the postfix from the link. Allthough the
postfix is added in the source read phase of the build, they must be removed
later, since a-plus uses combined chapterkeys.

The recommendable file structure in multilingual courses is to have corresponding
materials in the same folder, using identical names with postfixes. Some courses
use different ways of implementing the structure (for example traky), and this
change wont propably work with them all.

### DIFF
--- a/toc_config.py
+++ b/toc_config.py
@@ -370,6 +370,26 @@ def traverse_tocs(app, doc):
     names = []
     for toc in doc.traverse(addnodes.toctree):
         hidden = toc.get('hidden', False)
-        for _,docname in toc.get('entries', []):
-            names.append((docname,hidden))
-    return [(name,hidden,app.env.get_doctree(name)) for name,hidden in names]
+        for _, docname in toc.get('entries', []):
+            names.append((docname, hidden))
+    return [(name, hidden, app.env.get_doctree(name)) for name, hidden in names]
+
+
+# This function adds the language postfix to chapterlinks in multilingual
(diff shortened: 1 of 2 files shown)


toc_config.py
```diff
@@ -370,6 +370,31 @@ def traverse_tocs(app, doc):
...
+def add_lang_postfixes_to_links(app, docname, source):
+    # If the course use different format in links or for some other reason links
+    # need to stay untouched, add no_link_correction to conf.py or
+    # no-link-correction to the main index
+    if ("no_link_correction" not in app.env.metadata[app.config.master_doc]
+            and not app.config.no_link_correction):
+        postfix = docname[-3:]
+        if postfix[0] == '_':
+            # Links of the form :doc:`link text <path/file>` or `link text <path/file>`_
+            source[0] = re.sub(r"<([a-zA-Z0-9_/.]+/)?([a-zA-Z0-9_]+[^_]..)>",
```
alalav1: Now that I think of it, links of the form `link text <path/file>`_ shouldn't be touched at all, since they do not necessarily point to the course material, but in any page in general.

Mankro:
Is this PR replaced by the newer https://github.com/Aalto-LeTech/a-plus-rst-tools/pull/65?

alalav1:
This PR is replaced by the newer #65 which contains closely related changes.

PR #62, opened 2019-12-09
Enumerate questions

alalav1:
Description from Trello:
```
Currently, the feedback for  pick-any questions is provided in a bulleted list.
If a student receives feedback on multiple options, it can be difficult to identify which feedback refers to what choice.
```

This PR adds the enumerate option to pick-one and pick-any questions in questionnaires. 
The enumerate option has two possible values: numeric or alphabet.

### DIFF
--- a/directives/questionnaire.py
+++ b/directives/questionnaire.py
@@ -304,10 +304,14 @@ class QuestionMixin:
             if value.startswith('regexp:'):
                 compare_regexp = True
                 value = value[7:]
-            if enumeration_of_choices and not isnot:
-                line[0] = enumeration_of_choices[value] + content.strip()
-            else:
-                line[0] = content.strip()
+
+            line[0] = content.strip()
+
+            if enumeration_of_choices:
+                if value.startswith('%100%'):
+                    line[0] = enumeration_of_choices[value[5:]] + line[0]
+                else:
+                    line[0] = enumeration_of_choices[value] + line[0]
 
             # Create document elements.
             hint = aplus_nodes.html(u'div')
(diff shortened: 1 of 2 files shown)


alalav1:
The hint can be added when the answer option is selected or when it is not selected. This may be problematic. For example, if correct key is c the hint could be `!c § try to use fourier analysis` and the hint label would immediately show the correct answer. I suggest that if the hint has is marked as a negation, it wouldn't have the enumeration.

directives/questionnaire.py
```diff
@@ -303,6 +304,10 @@ def add_feedback(self, node, data, paragraph):
             if value.startswith('regexp:'):
                 compare_regexp = True
                 value = value[7:]
+            if enumeration_of_choices and not isnot:
+                line[0] = enumeration_of_choices[value] + content.strip()
+            else:
+                line[0] = content.strip()
```
jaguarfi, line 310: When the ``!`` symbol is prepended, the numeration should be still added. Otherwise, it is a bit annoying to have one line of feedback without the numeration.

jaguarfi:
The documentation in the Readme file is missing.

jaguarfi:
I accidentally pushed the changes to Vesa's branch. I will create a PR mentioning this PR and the changes implemented by Vesa.

Mankro:
I am closing this. This work has become outdated and the feature is not needed anymore. Since A+ v1.10, questionnaires may use the option `checkbox_feedback` in checkbox questions if they want to render the hint directly under the affected checkbox. This accomplishes a similar effect as the enumeration concerning how the hint is connected to each checkbox.

`checkbox_feedback`:
https://github.com/apluslms/mooc-grader/blob/ac73c32cef61d2cc7c425fccdb2bd0991d243b55/courses/README.md

PR #64, opened 2020-01-09
no_poi_box option in POI

zunde:
New option for point of interest (POI). With this option it is possible to hide the surrounding box from POI.

I created new css class (no_poi_box) to remove the indentation from POI to make it blend better with rest of the text.

Changes in point_of_interest.py: 
- new option (:no_poi_box:)
- if `:no_poi_box` option is active then css classes are selected differently and navigation is removed from POI.

### DIFF
--- a/directives/point_of_interest.py
+++ b/directives/point_of_interest.py
@@ -13,6 +13,7 @@ Directive for creating "point of interest" summary block.
     :columns: relative widths of poi content columns (e.e. 2 3 3)
     :not_in_slides: used with the presentation maker. This POI does not show in the slides when used.
     :not_in_book: This POI does not appear in the book material when used
+    :no_poi_box: Removes surrounding box and navigation
 
     Content of point-of-interest here
 
@@ -54,6 +55,7 @@ class PointOfInterest(Directive):
         # not_in_slides and not_in_book are used with the presentation maker
         'not_in_slides': directives.flag,
         'not_in_book': directives.flag,
+        'no_poi_box': directives.flag,
     }
 
     def run(self):
@@ -79,9 +81,6 @@ class PointOfInterest(Directive):
                         break
             title_text = self.arguments[0]
 
-        container_class = 'poi-container'
-        content_name = name + '-content'
-
         node = nodes.container()
         title = nodes.container()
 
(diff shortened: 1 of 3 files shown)


Mankro:
The new directive option must be described in the a-plus-rst-tools README. There is already a section for the point-of-interest directive.

Mankro:
You can squash your two git commits into one single commit.

directives/point_of_interest.py
```diff
@@ -90,11 +89,23 @@ def run(self):
...
+
+            # add an extra div to force content to desired height
+            hcontainer_opts = {
+                u'style': 'height:' + self.options.get('height', '') + ';',
+                u'class': container_class + ' poi-content row',
+            }
+        else:
+            container_class = 'no_poi_box'
+            content_name = name
+            hcontainer_opts = {
```
Mankro: No need to repeat identical `hcontainer_opts` definition twice. Define it outside the if statement since it is not affected by the if condition (the option `no_poi_box`).

directives/point_of_interest.py
```diff
@@ -90,11 +89,23 @@ def run(self):
...
+            container_class = 'poi-container'
+            content_name = name + '-content'
+
+            # add an extra div to force content to desired height
+            hcontainer_opts = {
+                u'style': 'height:' + self.options.get('height', '') + ';',
+                u'class': container_class + ' poi-content row',
+            }
+        else:
+            container_class = 'no_poi_box'
```
Mankro: The naming style for CSS classes is to use hyphen `-`, not underscore `_`. So the class name becomes `no-poi-box`.

directives/point_of_interest.py
```diff
@@ -54,6 +55,7 @@ class PointOfInterest(Directive):
         # not_in_slides and not_in_book are used with the presentation maker
         'not_in_slides': directives.flag,
         'not_in_book': directives.flag,
+        'no_poi_box': directives.flag,
```
Mankro: I would use hyphens `-` in option names instead of underscores `_`, but the existing `not_in_slides` option has already been named like that. I guess we keep underscores here.

directives/point_of_interest.py
zunde: It is affected by the if condition, compare the lines 99 and 106. In the else condition `u'class` has only `row` as a value.

directives/point_of_interest.py
zunde: I will fix this.

directives/point_of_interest.py
zunde: Okay.

PR #81, opened 2020-04-17
Column rows

zunde:
# Description
New column and row directives. Math roles can be used in titles. Relevant parts in the readme updated.

# Testing

- [ ] All unit tests pass
- [X] I have tested manually

- [X] Yes
- [ ] Not yet. I will do it next.
- [ ] Not relevant

### DIFF
--- a/directives/row.py
+++ b/directives/row.py
@@ -1,0 +1,120 @@
+"""
+Directive for creating nested multirow columns with Bootstrap.
+
+Designed to be used with the point-of-interest directive.
+
+Options
+:width: sets column width. Maximum value 12.
+:column-class: Bootstrap classes can be applied. See example below.
+
+More about the Bootstrap classes https://getbootstrap.com/docs/3.4/css/#helper-classes-backgrounds
+
+
+.. point-of-interest:: Test
+
+   .. row::
+
+     .. column::
+       :width: 8
+       :column-class: bg-warning text-center
+
+       .col-8 this is column's content.
+
+       .. row::
+
+         .. column::
+           :width: 6
+           :column-class: bg-light
+
+           .col-6
+
+         .. column::
+           :width: 6
+           :column-class: bg-secondary
+
+           .col-6
+
+     .. column::
+       :width: 4
+       :column-class: bg-success
+
+       .col-4
+
+"""
+
+from docutils.parsers.rst import Directive, directives
+from docutils import nodes
+
+import aplus_nodes
+
+
+class ColumnNode(nodes.General, nodes.Element):
+
+    def __init__(self):
+        super(ColumnNode, self).__init__()
+
+    @staticmethod
+    def visit_column(self, node):
+        self.body.append(self.starttag(node, 'div'))
+
+    @staticmethod
+    def depart_column(self, node=None):
+        self.body.append('</div>\n')
+
+
+class Column(Directive):
+    option_spec = {
+        'width': directives.unchanged,
+        'column-class': directives.unchanged,
+    }
+
+    has_content = True
+
+    def run(self):
+        self.assert_has_content()
+        node = ColumnNode()
+        if 'width' in self.options:
+            # Bootstrap 3 does not support automatic widths in columns.
+            col_width = str(self.options['width']).strip()
+            if not col_width:
+                col_width = ""
+            else:
+                col_width = "-" + col_width
+        else:
+            col_width = ""
+        node['classes'].append('col-sm' + col_width)
+
+        if 'column-class' in self.options:
+            column_classes = [c.strip() for c in str(self.options['column-class']).split()]
+            column_classes = [c for c in column_classes if c]
+            if column_classes:
+                node['classes'].extend(column_classes)
+
+        self.state.nested_parse(self.content, self.content_offset, node)
+        return [node]
+
+
+class Row(Directive):
+
+    has_content = True
+
+    def run(self):
+        self.assert_has_content()
+        container_class = 'row'
+        row_container_opts = {
+            u'class': container_class,
+        }
+
+        row_container = aplus_nodes.html(u'div', row_container_opts)
+        self.state.nested_parse(self.content, self.content_offset, row_container)
+
+        return [row_container]
+
+
+def setup(app):
+    ignore_visitors = (aplus_nodes.visit_ignore, aplus_nodes.depart_ignore)
+    app.add_directive('row', Row)
+
+    app.add_node(ColumnNode, html=(ColumnNode.visit_column, ColumnNode.depart_column),
+                 latex=ignore_visitors)
+    app.add_directive('column', Column)
--- a/README.md
+++ b/README.md
@@ -713,6 +713,7 @@ Point of interests may also be used to generate separate lecture slides
 (not directly included in the A+ content chapters). This requires a separate
 tool called "presentation maker".
 
+More information about [Columns and rows](#17-columns-and-rows).
 ```
 .. point-of-interest:: Title text
   :id: unique id, if not supplied a random id will be generated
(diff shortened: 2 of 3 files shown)


directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+       :width: 8
+       :column_class: bg-warning text-center
+
+       .col-8 this is column's content.
+
+       .. row::
+
+         .. column::
+           :width: 6
+           :column_class: bg-light
```
raphendyr, line 27: The key here is not updated to `column-class`

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+
+           .col-6
+
+         .. column::
+           :width: 6
+           :column_class: bg-secondary
```
raphendyr, line 33: The key here is not updated to `column-class`

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+
+           .col-6
+
+     .. column::
+       :width: 4
+       :column_class: bg-success
```
raphendyr, line 39: The key here is not updated to `column-class`

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+More about the Bootstrap classes https://getbootstrap.com/docs/4.4/layout/grid/#row-columns
+
+
+.. point-of-interest:: Test
+
+   .. row::
+
+     .. column::
+       :width: 8
+       :column_class: bg-warning text-center
```
raphendyr, line 19: The key here is not updated to `column-class`

directives/row.py
```diff
@@ -0,0 +1,126 @@
+"""
+Directive for creating nested multirow columns with Bootstrap.
+
+Designed to be used with the point-of-interest directive.
+
+Options
+:width: sets column width. Maximum value 12.
+:column_class: Bootstrap classes can be applied. See example below.
```
raphendyr, line 8: The key here is not updated to `column-class`

raphendyr:
Introduces a backward incompatible change: removes `::newcol`. Thus, raises major version number or there should be a backward compatibility feature.

raphendyr:
A side note: When you have a commit message including multiple actions (e.g. there is a comma), then it's typically good idea to split it to multiple commits. `git add` and `git add -p` might help there.

Another side note: after feature implementetaion, it would be a good time to revisit commits and clean them up. E.g. when adding a name and then renaming it in a later commit, could be rebased to a single commit.

raphendyr:
From my point of view `:width:` seems a loosely defined argument. By default, I think it should be automatically calculated so that cols wihtin a row would be evenly distributed. In addition, I think the value shouldn't be in range of 1 to 12, but e.g. 50% or 75% and so on. In other words, this introduces strong link to specific version of a bootstrap. As it is now, I don't think it adds much value compared to not having it, which would require manually writing classes like `col-xs-3`.

Interestingly, is there many other locations in rst tools, that make strong assumptions about the used CSS classes in A+ (e.g. bootstrap version)?

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+            if not col_width:
+                # width is empty.
+                col_width = ""
+            else:
+                # not empty
+                col_width = "-" + col_width
+        else:
+            col_width = ""
+
+        if 'column-class' in self.options:
```
raphendyr, line 87: What you think about if this would be written:

```
if 'column-class' in self.options:
  column_classes = [c.strip() for c in str(self.options['column-class']).split()]
  column_classes = [c for c in column_classes if c]
  if column_classes:
    node['classes'].extend(column_classes)
```

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+
+from docutils.parsers.rst import Directive, directives
+from docutils import nodes
+
+import aplus_nodes
+
+
+class ColumnNode(nodes.General, nodes.Element):
+
+    def __init__(self, text):
```
Mankro, line 53: Why does the node constructor have the unused `text` argument? Since it is not needed, could you remove it?

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+        if 'column-class' in self.options:
+            column_class = str(self.options['column-class'])
+            if not column_class:
+                column_class = ""
+            else:
+                column_class = " " + column_class
+        else:
+            column_class = ""
+        column_content = '\n'.join(self.content)
+        node = ColumnNode(column_content)
```
Mankro, line 96: If the text argument is removed from the `ColumnNode` constructor, then you would not supply the parameter here either.

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+    def depart_column(self, node=None):
+        self.body.append('</div>\n')
+
+
+class Column(Directive):
+    option_spec = {'width': directives.unchanged,
+                   'column-class': directives.unchanged,
+                   }
+
+    final_argument_whitespace = True
```
Mankro, line 70: Is `final_argument_whitespace = True` really needed? It looks like the directive does not use any arguments at all. If the default False value is OK, then you don't need to define it at all. 
https://www.sphinx-doc.org/en/master/extdev/markupapi.html#docutils.parsers.rst.Directive.final_argument_whitespace

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+    def visit_column(self, node):
+        self.body.append(self.starttag(node, 'div'))
+
+    @staticmethod
+    def depart_column(self, node=None):
+        self.body.append('</div>\n')
+
+
+class Column(Directive):
+    option_spec = {'width': directives.unchanged,
```
Mankro, line 66: Better coding style:

```
option_spec = {
    'width': directives.unchanged,
    'column-class': directives.unchanged,
}
```

directives/row.py
```diff
@@ -0,0 +1,126 @@
...
+            column_class = ""
+        column_content = '\n'.join(self.content)
+        node = ColumnNode(column_content)
+        node['classes'].append('col-sm' + col_width + column_class)
+        self.state.nested_parse(self.content, self.content_offset, node)
+        return [node]
+
+
+class Row(Directive):
+    required_arguments = 0
```
Mankro, line 103: `required_arguments = 0` is the default value so you don't need to define it again.

directives/row.py
```diff
@@ -0,0 +1,126 @@
+"""
+Directive for creating nested multirow columns with Bootstrap.
+
+Designed to be used with the point-of-interest directive.
+
+Options
+:width: sets column width. Maximum value 12.
+:column_class: Bootstrap classes can be applied. See example below.
+
+More about the Bootstrap classes https://getbootstrap.com/docs/4.4/layout/grid/#row-columns
```
Mankro, line 10: The link refers to Bootstrap 4 but A+ uses the version 3. We can't easily upgrade the version, thus we must stay at version 3 for the time being.

Mankro:
- [ ] there are reviewer's comments in some parts of the code that you should fix
- [ ] git commit history needs to be cleaned up. The commit messages must be formatted well. The first line is the title and it should not be longer than 70 characters. I think that all commits could be squashed into one.
- [ ] `::newcol` was supposed to be still supported, but you have removed it. I would say that the RST source is not allowed to mix ::newcol with the new column/row directives, but old content with ::newcol should still work like it used to. Of course, the POI content may also be just plain content without any columns (no ::newcol nor new column directives).

Jaakko mentioned the column width option. It is indeed a little awkward to manually count to 12, but Bootstrap is not very flexible with this. If we used percentages, then they would be rounded a lot in some cases since the underlying Bootstrap framework still uses 12 columns. It would not exactly match the given percentages. I think we need to use the numbers 1-12 here. However, you could add automatic default widths so that if there are no width, then all columns are set to the same width. The width would depend on the number of columns. Is this too hard to implement since you would need to count the number of columns before you could the width on a column?

Mankro:
Just to emphasize, the `::newcol` syntax should be supported in the a-plus-rst-tools for the sake of the old courses in A+, but the Presentation maker project does not need to support `::newcol`.

PR #82, opened 2020-04-23
Add the js and css implementation of the annotated directive.

jaguarfi:
# Description
The annotated directive has been part of the a-plus-rst tools for a
long time. However, the annotated directive required a custom
implementation of JS and CSS. Therefore, a default implementation is
added to make this directive easier to use.

It is important to note that this default implementation can be omitted
by setting the 'include_annotated_js' and 'include_annotated_cs'
parameters to False inside the conf.py file.

Side note: If the 'include_annotated_js' and 'include_annotated_cs' are
set to False or omitted from the conf.py, the user should create and add
their own CSS and JS implementation for the annotated directive.

# Testing

- [ ] All unit tests pass
- [X] I have tested manually

<sub>First test case</sub>
annotated
<sub>Second test case</sub>
annotated2

- [X] Yes
- [ ] Not yet. I will do it next.
- [ ] Not relevant

### DIFF
--- a/directives/annotated.py
+++ b/directives/annotated.py
@@ -8,9 +8,18 @@ import re
 import os
 from sphinx.directives.code import CodeBlock
 from sphinx.errors import SphinxError
+from sphinx.util.fileutil import copy_asset
+from sphinx.util import logging
 from operator import itemgetter
 
 import aplus_nodes
+
+CSS_FILE = 'css/annotated.css'
+JS_FILE = 'js/annotated.js'
+
+assets_path = 'static'
+
+logger = logging.getLogger(__name__)
 
 annotated_section_counts = Counter()
 
@@ -357,10 +366,41 @@ def annotate(html, section_name, annotations):
     content = u''.join(result)
     return u''.join([parts[0], parts[1], content, parts[3], parts[4]])
 
+def add_assets(app):
+    # This method reads the `include_annotated_css` and `include_annotated_js`
+    # settings from the conf.py file located in the course directory. If such
+    # settings are not found, the default settings defined in the setup()
+    #  method will be used instead
+    app.config.include_annotated_css and app.add_stylesheet(CSS_FILE)
+    app.config.include_annotated_js and app.add_javascript(JS_FILE)
 
+def copy_asset_files(app, exc):
+
+    if exc:
+        return
+        
+    # The files are added to the _build/html/_static/css folder. 
+    if app.config.include_annotated_css:
+        logger.info('Copying CSS files from the annotated directive to the _static folder... ')
+        html_static_path_css = os.path.join(assets_path, CSS_FILE)
+        local_path_css = os.path.join(os.path.dirname(__file__), html_static_path_css)
+        copy_asset(local_path_css, os.path.join(app.outdir, '_static', 'css'))
+        logger.info('done')
+    
+    # The files are added to the _build/html/_static/js folder. 
+    if app.config.include_annotated_js:
+        logger.info('Copying JS files from the annotated directive to the _static folder... ')
+        html_static_path_js = os.path.join(assets_path, JS_FILE)
+        local_path_js = os.path.join(os.path.dirname(__file__), html_static_path_js)
+        copy_asset(local_path_js, os.path.join(app.outdir, '_static', 'js'))
+        logger.info('done')
+        
 def setup(app):
 
     ignore_visitors = (aplus_nodes.visit_ignore, aplus_nodes.depart_ignore)
+
+    app.add_config_value('include_annotated_css', False, 'html')
+    app.add_config_value('include_annotated_js', False, 'html')
 
     app.add_node(annotated_node, html=(visit_annotated_node, depart_annotated_node),
             latex=ignore_visitors)
@@ -373,3 +413,7 @@ def setup(app):
     app.add_node(altered_node, html=(visit_altered_node, depart_altered_node),
             latex=ignore_visitors)
     app.add_directive('altered-code-block', AlteredCodeBlock)
+
+    app.connect('builder-inited', add_assets)
+
+    app.connect('build-finished', copy_asset_files)
--- a/theme/aplus/layout.html
+++ b/theme/aplus/layout.html
@@ -37,6 +37,14 @@
     <script type="text/javascript" async src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML"></script>
     {%- endif %}
 
+    {% if '_static/css/annotated.css' in css_files %}
+    <link rel="stylesheet" href="{{ pathto('_static/css/annotated.css', 1) }}" type="text/css" data-aplus="yes"/>
+    {% endif %}
+
+    {% if '_static/js/annotated.js' in script_files %}
+    <script rel="stylesheet" src="{{ pathto('_static/js/annotated.js', 1) }}" type="text/javascript" data-aplus="yes"></script>
+    {% endif %}
+
     <link rel="stylesheet" href="https://plus.cs.hut.fi/static/css/main.css" />
 
     {%- if style %}
--- a/README.md
+++ b/README.md
@@ -741,11 +741,24 @@ tool called "presentation maker".
 ### 10. Annotated code blocks
 
 Code blocks may be annotated with comments for specific lines. This extension
-must be activated separately in the project conf.py
-(`extensions = ["aplus_setup", "annotated"]`).
-This extension requires custom JavaScript code and CSS styles in order to
-highlight the annotations on mouse hover in the web browser. The frontend code
-is not distributed in this repository (or anywhere).
+must be activated separately in the project by adding the following settings to
+the **conf.py** file located in the root of your course directory.
+
+```python
+extensions = ["aplus_setup", "annotated"]
+...
+include_annotated_js = True
+include_annotated_css = True
+```
+
+This directive requires a custom JavaScript and CSS implementation for
+highlighting the annotated code when the mouse hover events of the annotations
+are fired in the web browser. Therefore, we have added a default implementation
+that allow you to make use of the annotated directive without having to write
+your own JS or CSS. However, if you want to remove the default JavaScript and
+CSS implementation, you can change the value of `include_annotated_js` and
+`include_annotated_css` to `False`. By doing so, you could write your own CSS
+and JS code to interact with the annotated directive.
 
 ```
 .. annotated::
--- a/directives/static/css/annotated.css
+++ b/directives/static/css/annotated.css
@@ -1,0 +1,107 @@
+/* hover highlights for the annotated a-plus-rst directive */
+.content div.codecomment,
+.content div.container.guicomment {
+  border-radius: 7px;
+  -webkit-border-radius: 7px;
+  -khtml-border-radius: 7px;
+  -moz-border-radius: 7px;
+  border-style: solid;
+  border-width: 1px 0px 1px 0px;
+  width: 80%;
+  padding: 10px 15px 10px 15px;
+  margin: 0px 12px 12px 95px;
+  background: #d9eecf;
+  box-shadow: 2px 2px 2px 0px #bfc2a1;
+  border-color: #bfc2a1;
+  font-size: 0.9em;
+  line-height: 1.5;
+}
+.content span.codecomment {
+  border-radius: 7px;
+  -webkit-border-radius: 7px;
+  -khtml-border-radius: 7px;
+  -moz-border-radius: 7px;
+  border-style: solid;
+  border-width: 1px 0px 1px 0px;
+  margin: 0px 1px 0px 1pxs;
+  padding: 0px 2px 1px 2px;
+  background: #d9eecf;
+  box-shadow: 1px 1px 1px 0px #bfc2a1;
+  border-color: #bfc2a1;
+  white-space: nowrap;
+}
+.content div.admonition.aside div.codecomment {
+  width: 90%;
+  font-size: 0.85em;
+}
+.content div.codecomment + p,
+.content div.guicomment + p {
+  margin-top: 18px;
+}
+.content div.codecomment.comment-now-highlighted,
+.content div.guicomment.comment-now-highlighted {
+  box-shadow: 0px 0px 3px 3px #1ea780;
+}
+.content span.codecomment.comment-now-highlighted {
+  box-shadow: 0px 0px 2px 2px #1ea780;
+}
+.content span.loc-now-highlighted,
+.content div.highlight pre span.loc-now-highlighted,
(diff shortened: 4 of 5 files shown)


directives/annotated.py
```diff
@@ -74,7 +79,10 @@ def run(self):
 
     def assert_sanity(self, content):
         annotation_numbers_present = set(map(lambda matching: int(matching[0]), re.findall(u"\d«", content)))
-        highest_present = max(annotation_numbers_present)
+        if len(annotation_numbers_present) == 0:
+            highest_present = 1
```
Mankro, line 83: Why is this `if` necessary?

directives/annotated.py
```diff
@@ -357,6 +365,22 @@ def annotate(html, section_name, annotations):
...
     return u''.join([parts[0], parts[1], content, parts[3], parts[4]])
 
+def add_assets(app):
+    app.add_stylesheet(CSS_FILE)
+    app.add_javascript(JS_FILE)
+
+def copy_assets(app, exception):
+    if app.builder.name not in ['html', 'readthedocs'] or exception:
+        return
+    logger = logging.getLogger(__name__)
```
Mankro, line 375: Usually, we initialize the logger at the start of the module, not inside a function.

directives/annotated.py
```diff
@@ -357,6 +365,22 @@ def annotate(html, section_name, annotations):
...
+    logger.info('Copying aplus annotated stylesheet/javascript... ', nonl=True)
+    dest = os.path.join(app.builder.outdir, '_static', CSS_FILE)
```
Mankro, line 377: The directory name `_static` should not be hardcoded since it can be modified in the `conf.py` file. You should be able to read the value from `app.config` or something like that.

directives/annotated.py
```diff
@@ -373,3 +397,6 @@ def setup(app):
     app.add_node(altered_node, html=(visit_altered_node, depart_altered_node),
             latex=ignore_visitors)
     app.add_directive('altered-code-block', AlteredCodeBlock)
+
+    app.connect('builder-inited', add_assets)
```
Mankro, line 401: I thought that `add_assets` could be called directly in `setup` and using the `builder-inited` signal is unnecessary.

directives/css/annotated.css
```diff
@@ -0,0 +1,72 @@
+/* hover highlights for the annotated aplus-rst directive ------------------------------------------------------------------------ */
```
Mankro, line 1: Why is the comment so long? All that `---`

Mankro:
Some courses already include the annotated JS and CSS code via the course static files. Those courses could be broken when the aplus-rst-tools include them again. The a-plus-rst-tools should have an option for not including the CSS/JS files. The Sphinx HTML theme options are available in the HTML template as variables.

For example, see the old include_mathjax setting:
https://github.com/Aalto-LeTech/a-plus-rst-tools/blob/d2dd81fa98a5e7098180de3cc7f8e0c6f134e6df/theme/aplus/layout.html#L36
https://github.com/Aalto-LeTech/a-plus-rst-tools/blob/d2dd81fa98a5e7098180de3cc7f8e0c6f134e6df/theme/aplus/theme.conf#L9

https://www.sphinx-doc.org/en/master/theming.html#creating-themes
https://www.sphinx-doc.org/en/master/templating.html#script_files

directives/annotated.py
```diff
@@ -357,6 +365,22 @@ def annotate(html, section_name, annotations):
     content = u''.join(result)
     return u''.join([parts[0], parts[1], content, parts[3], parts[4]])
 
+def add_assets(app):
+    app.add_stylesheet(CSS_FILE)
+    app.add_javascript(JS_FILE)
+
+def copy_assets(app, exception):
```
Mankro, line 372: Is it really necessary to manually copy the static JS/CSS files like this?

Can we use the static directory of the a-plus-rst-tools HTML theme for the CSS and JS files?
https://github.com/Aalto-LeTech/a-plus-rst-tools/tree/master/theme/aplus/static
Is JS added with `app.add_javascript()` available in the Sphinx HTML template in the `script_files` template variable?
https://github.com/Aalto-LeTech/a-plus-rst-tools/blob/d2dd81fa98a5e7098180de3cc7f8e0c6f134e6df/theme/aplus/layout.html#L55

https://www.sphinx-doc.org/en/master/theming.html#creating-themes
https://www.sphinx-doc.org/en/master/templating.html#script_files

Mankro: If you use the static directory of the a-plus-rst-tools HTML theme, I think that the course repos do not need to manually add the JS/CSS files in any course-specific `_templates/layout.html` template.

directives/annotated.py
jaguarfi: It is is not necessary, it is good that you saw this change. I just forgot to remove this line that I added when I was trying something.

directives/annotated.py
jaguarfi: I think this _static folder is always created by Sphinx, and therefore, it should be harcoded.
https://github.com/sphinx-doc/sphinx/issues/2202

directives/annotated.py
jaguarfi: I added a comment explaining why

directives/annotated.py
Mankro: Right, seems like you are correct. So the build output directory may use a hardcoded `_static` directory while the static "source" directory under the course repo may be renamed in `conf.py`.

directives/annotated.py
Mankro: Files from the theme static directory are always copied to the build output static directory. On the contrary, using custom code for copying the JS and CSS files to the build static directory enables more control: the JS and CSS do not need to be always copied. We don't need to copy the annotated JS and CSS files if the course does not use the annotated directive at all or if the course uses its own versions of the files in its `_static` directory.

directives/annotated.py
Mankro: I agree now that the signals are required. The comment is important for explaining it.

Mankro (changes requested):
Overall, this is good. Some details need to be fixed and I don't want to include unrelated coding style changes in this PR. There were so many of those coding style changes that they hide the real changes in the git diff.

Mankro:
@jaguarfi Please update the pull request description since it is now outdated and incorrect.

directives/annotated.py
```diff
@@ -8,78 +8,102 @@
...
 annotated_section_counts = Counter()
 
+
 class AnnotationError(SphinxError):
     category = 'Annotation error'
 
 
 def clean_path(path):
-  return re.sub(r"[/\\ :]+", "", path).replace(".rst", "")
+    return re.sub(r"[/\\ :]+", "", path).replace(".rst", "")
```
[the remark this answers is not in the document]
jaguarfi: I will remove all the changes in style.

directives/annotated.py
```diff
@@ -358,18 +416,59 @@ def annotate(html, section_name, annotations):
...
+def copy_asset_files(app, exc):
+    if exc is None:
+        logger.info(
+            'Copying css files from the annotated directive to the _static folder... ')
+        # The files are added to the _build/html/_static/css folder.
+        if app.config.aplus_rst_tools_annotated_css:
+            html_static_path_css = os.path.join(assets_path, CSS_FILE)
+            local_path_css = resource_filename(__name__, html_static_path_css)
+            copy_asset(local_path_css, os.path.join(
+                app.outdir, assets_path, 'css'))
```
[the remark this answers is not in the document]
jaguarfi: I created the `assets_path` variable, and I am using it for storing the value `_static` which is used as the hardcoded Sphinx directory name.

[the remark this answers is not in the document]
jaguarfi: I created the assets_path variable, and I am using it for storing the value _static which is used as the hardcoded Sphinx directory name.

jaguarfi:
It took me a considerable amount of time to remove some pieces of code that were part of the commits, test the implementation and remember what I was trying to accomplish with this PR, but I think that now, it is ready to be reviewed one more time.

theme/aplus/layout.html
```diff
@@ -48,13 +48,13 @@
...
-    {%- endfor %#}
+    {%- for cssfile in css_files %}
+    <link rel="stylesheet" href="{{ pathto(cssfile, 1) }}" type="text/css" data-aplus="yes"/>
+    {%- endfor %}
 
-    {#%- for scriptfile in script_files %}
-    <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
-    {%- endfor %#}
+    {%- for scriptfile in script_files %}
+    <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}" data-aplus="yes"></script>
```
Mankro, line 56: It looks like that `css_files` and `script_files` contain plenty of files included by Sphinx that we don't want to have in A+. That is why those variables used to be commented out. With your change, all those files are included with the `data-aplus` attribute and we can't have that. They might interfere with A+ or some course contents. Some of the JS even crashes on every page load (I can see errors in the browser console). 

Can the template (layout.html) use conf.py variables in if statements in order to avoid the `script_files` variable? If yes, the `include_annotated_js` variable could be used in the template too.

Mankro (changes requested):
I found a problem in the use of `script_files` in the template, see below.

The annotated.js seems to be an old version that does not support inline annotations. The README describes the syntax for inline annotations. Did you test them at all? Inline annotations also support replacement text and we want to support that too. The Programming 1 course has the version of the JS and CSS code that supports inline annotations.

In the rendered web page, the annotation box has very low contrast between the text colour and the background. This needs to be fixed. The WAVE browser addon warned about it.

Is there a reason to separate the README git commit from the code commit? I think they can be squashed into one commit.

theme/aplus/layout.html
Mankro: If the conf.py variables can not be directly used in the template, then at least the theme settings can be and they are also defined in conf.py (in addition to the `theme/aplus/theme.conf` file that I mentioned in the comments in April). In that case, we would just need to use the theme settings dictionary in conf.py and the annotated.py code would check the flag from there. The `add_javascript_file` and `add_css_file` functions would not be needed anymore.

jaguarfi: Yes, I will use the theme options. It seems more logic to add those options to the theme than creating isolated variables in the template.

jaguarfi:
> The annotated.js seems to be an old version that does not support inline annotations. The README describes the syntax for inline annotations. Did you test them at all? Inline annotations also support replacement text and we want to support that too. The Programming 1 course has the version of the JS and CSS code that supports inline annotations.

This is interesting, who document features only in the code examples?. Anyway, I did not test that feature. But now, it has been tested and it is working.
> 
> In the rendered web page, the annotation box has very low contrast between the text colour and the background. This needs to be fixed. The WAVE browser addon warned about it.

Yes, it is true. I added those styles a long time ago, and I did not take into account the accessibility. Now, the contrast should be fine.
> 
> Is there a reason to separate the README git commit from the code commit? I think they can be squashed into one commit.

Once the changes are accepted, I will squash all the commits.

theme/aplus/theme.conf
```diff
@@ -7,3 +7,5 @@ pygments_style = colorful
 nosidebar = True
 use_wide_column = True
 include_mathjax = True
+include_annotated_js = True
```
Mankro, line 10: Let's use `False` as the default for including annotated JS/CSS.

directives/annotated.py
```diff
@@ -357,11 +366,42 @@ def annotate(html, section_name, annotations):
...
+        html_static_path_js = os.path.join(assets_path, JS_FILE)
+        local_path_js = os.path.join(os.path.dirname(__file__), html_static_path_js)
+        copy_asset(local_path_js, os.path.join(app.outdir, '_static', 'js'))
+        logger.info('done')
+        
 def setup(app):
 
     ignore_visitors = (aplus_nodes.visit_ignore, aplus_nodes.depart_ignore)
 
+    app.add_config_value('include_annotated_css', False, 'html')
```
Mankro, line 402: `app.add_config_value` should not be used anymore after you switched to the theme options.

directives/annotated.py
```diff
@@ -357,11 +366,42 @@ def annotate(html, section_name, annotations):
     content = u''.join(result)
     return u''.join([parts[0], parts[1], content, parts[3], parts[4]])
 
-
+def add_assets(app):
+    # This method reads the `include_annotated_css` and `include_annotated_js`
+    # settings from the conf.py file located in the course directory. If such
+    # settings are not found, the default settings defined in the setup()
+    #  method will be used instead
+    app.config.include_annotated_css and app.add_stylesheet(CSS_FILE)
```
Mankro, line 374: You are using the old config variable, but you switched to theme options. In addition, `app.add_stylesheet` is not needed with the theme options. The same for JS.

directives/annotated.py
```diff
@@ -357,11 +366,42 @@ def annotate(html, section_name, annotations):
...
+    app.config.include_annotated_js and app.add_javascript(JS_FILE)
+
+def copy_asset_files(app, exc):
+
+    if exc:
+        return
+        
+    # The files are added to the _build/html/_static/css folder. 
+    if app.config.include_annotated_css:
```
Mankro, line 383: The old config variable shall be replaced with the theme option.

Mankro (changes requested):
You didn't update the new theme config variable in `directives/annotated.py` yet. And the default in `theme/aplus/theme.conf` should be `False`. Otherwise, it is good.

jaguarfi:
I decided not to use the theme for adding the static files. I think that the directives and any related static file should be kept separate from the theme. I also think that if we keep the changes proposed in this PR and eventually decide to migrate to a newer version of Sphinx, we could use the new methods [add_css_file](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file) and the [add_js_file](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file). These new methods will allow us to add the attribute `data-aplus` and use the code already existing in the `layout.html`
```html
    {#%- for cssfile in css_files %}
    <link rel="stylesheet" href="{{ pathto(cssfile, 1) }}" type="text/css" />
    {%- endfor %#}

    {#%- for scriptfile in script_files %}
    <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
    {%- endfor %#}
```

conf.py
```diff
@@ -135,7 +135,7 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    'use_wide_column': use_wide_column,
+    'use_wide_column': use_wide_column
```
Mankro, line 138: No need to remove this comma.

Mankro (changes requested):
It's good now. You should return that one comma in conf.py and then squash the commits.

Mankro (approved):

PR #83, opened 2020-04-23
Tabs

jaguarfi:
# Description
Add tab component which allows navigation through different views.
Each view can contain multiple directives or roles inside.

Tabs

I created this directive to switch between RST implementation and the HTML render. Nevertheless, I think this implementation can be used to compare implementation in different programming languages.
For Example:
image

This pull request Includes the changes in the annotated directive!!!

# Testing

- [ ] All unit tests pass
- [X] I have tested manually

- [ ] Yes
- [X] Not yet. I will do it next.
- [ ] Not relevant


directives/_static/css/aplus_tabs.css
```diff
@@ -0,0 +1,71 @@
+@import url("default.css");
```
Mankro, line 1: I don't think this file should import the CSS styles of the whole aplus theme. Isn't the purpose of this aplus_tabs.css file to only contain the rules for the content tabs and these styles are only included in the A+ chapter if tabs are used there?

directives/_static/js/aplus_tabs.js
```diff
@@ -0,0 +1,45 @@
+$(function() {
+
+    $('div.content-tabs').each(function() {
+        var contenttab_sel = $('<ul />', { class: "contenttab-selector" });
```
Mankro, line 4: Our new JS style guide uses two spaces for indentation: https://apluslms.github.io/contribute/styleguides/

Mankro: `contenttab_sel` is actually a new element, not a selector ("sel" is abbreviated from "selector"). I'm also confused of the CSS class name "contenttab-selector". Why is the word "selector" used there? Does the name describe the purpose of the element?

directives/_static/js/aplus_tabs.js
```diff
@@ -0,0 +1,45 @@
...
+
+        var sel_class = $(this).attr('class');
+        $('div.contenttab',tabsblock).hide();
+        $('div#' + sel_class,tabsblock).show();
+
+        $('ul.contenttab-selector li', tabsblock).removeClass('selected');
+        $('ul.contenttab-selector li.' + sel_class, tabsblock).addClass('selected');
+
+        sel_class = null;
+    });
```
Mankro, line 42: Is this tab component different from the similar component in Bootstrap3? It looks like you implemented this tab component without Bootstrap. Is that correct? Why did you decide to avoid BS?

One reason that I can guess is that we can't easily control the HTML output of the RST compilation and that HTML structure might be very different from the one expected by the Bootstrap API.

directives/tabs.py
```diff
@@ -0,0 +1,129 @@
...
+import aplus_nodes
+
+CSS_FILE = 'css/aplus_tabs.css'
+JS_FILE = 'js/aplus_tabs.js'
+assets_path = '_static'
+
+logger = logging.getLogger(__name__)
+
+
+class ContentTabsDirective(Directive):
```
Mankro, line 46: The class names feel unintuitive to me at first. I would think that a "container" contains all the tabs while one "tab" is just the tab. That's just me. Is this all copied from the external library (sphinxcontrib-contentui)?

directives/tabs.py
```diff
@@ -0,0 +1,129 @@
...
+            local_path_js = resource_filename(__name__, html_static_path_js)
+            copy_asset(local_path_js, os.path.join(
+                app.outdir, assets_path, 'js'))
+
+        logger.info('done')
+
+
+def setup(app):
+
+    app.add_config_value('aplus_rst_tools_tabs_css', True, 'html')
```
Mankro, line 122: This is a new directive, so we don't have any old courses that would already have this JS/CSS code included elsewhere. Is it then necessary to have conf.py variables for disabling the JS/CSS?

directives/tabs.py
```diff
@@ -0,0 +1,129 @@
...
+    app.add_config_value('aplus_rst_tools_tabs_js', True, 'html')
+
+    app.add_directive('content-tabs',  ContentTabsDirective)
+    app.add_directive('tab-container', ContentTabsContainerDirective)
+
+    app.connect('builder-inited', add_assets)
```
Mankro, line 128: Using these signals is still a method for avoiding copying these files to the course build output directory if this tab directive is not used at all. Thus, I think we still want to use these signals instead of placing the JS/CSS files to the aplus theme static directory. That static directory would always be copied to the build output directory.

directives/tabs.py
```diff
@@ -0,0 +1,129 @@
...
+# Files added after triggering the 'builder-inited' event, in order to read the
+# settings from the conf.py file and prevent from reading the settings that
+# were added in the setup function
+def add_assets(app):
+    app.config.aplus_rst_tools_tabs_css and app.add_stylesheet(CSS_FILE)
+    app.config.aplus_rst_tools_tabs_js and app.add_javascript(JS_FILE)
+
+
+def copy_asset_files(app, exception):
+    if exception is None:
```
Mankro, line 95: I already commented on this function in the annotated PR #82 and some comments apply here too.

Mankro (changes requested):
This will be a useful component. Detailed comments below.

directives/_static/css/aplus_tabs.css
jaguarfi: Yes, I think I was doing some testing and I forgot to remove that import

directives/_static/js/aplus_tabs.js
jaguarfi: I am referring to the tab itself and not the content. So basically if I have a group of tabs, each tab is a selector for the content. But, I can make the names semantically more correct.

directives/_static/js/aplus_tabs.js
jaguarfi: Yes, I did not include any bootstrap here since I wanted to have an independent HTML element, able to work without relying on external CSS frameworks. I think it is better to keep the directives as independent as possible.

directives/tabs.py
jaguarfi: Continuing on https://github.com/apluslms/a-plus-rst-tools/pull/83#discussion_r500170379. The answer is yes. The concept here is similar to the concept utilised by bootstrap. They basically have the navigation tabs, and then they associate each tab to certain content by using ids. https://getbootstrap.com/docs/3.3/javascript/#markup

directives/tabs.py
jaguarfi: I think, that having this option allows the teachers to design their own Tabs. Maybe they would not want to change the JS code, but at least the CSS is something that they would like to change.

jaguarfi:
I had to force push the PR because it was behind the latest commits.

Mankro (changes requested):
This is nice. I have the following requests.

* Please add a description of the new tab directive to the README. 
* I tried this in the aplus-manual. The tab container was wider than the rest of the content in the page. It looked like that the CSS class `.container` was affected by the Bootstrap CSS in _grid.scss. This could maybe be fixed by adding some CSS rules to counter the Bootstrap defaults. The container class is used a lot in both Bootstrap and Sphinx and they are not compatible with each other.
* Is it intentional that all tabs can be hidden by clicking on the tab title multiple times?

jaguarfi:
I added the description of this new element to the **README.md** file. I also added a new CSS property to manage the width of the `tab-content` Regarding hiding the tab content when the user clicks on the current tabs, it is intentional behaviour. I think that such behaviour may help the user to save some space on the screen.

README.md
```diff
@@ -1015,7 +1015,56 @@ Width is not mandatory, but if it is not given then it uses the width of 12 auto
 
 **Note - newcol is deprecated**
 
-Older columns (`::newcol`) work, but they are deprecated. Column and row directives should be used instead.  
+Older columns (`::newcol`) work, but they are deprecated. Column and row directives should be used instead.
+
+## 18. Tabs
```
Mankro, line 1020: Other directives used level 3 headings (`### 18. Tabs`)

README.md
```diff
@@ -1015,7 +1015,56 @@ Width is not mandatory, but if it is not given then it uses the width of 12 auto
...
+
+Directive for adding tabbed content. It means that we can use tabs to separate content into different panels , allowing
+each panel to be viewed one at a time.
+
+The following snippet of code is an example on how you can use `aplus-tabs` directive. The `aplus-tabs` directive can
+contain one or more `tab-content` directives, and each `tab-content` takes an argument that is used internally to
+identify the content of each tab. Therefore, this argument must be unique and cannot contain whitespaces. You must also
+add the option `title` to the `tab-content` directive since it is the title that will be show in your tabs. The content
```
Mankro, line 1028: typo `show` -> `that will be shown`

README.md
```diff
@@ -1015,7 +1015,56 @@ Width is not mandatory, but if it is not given then it uses the width of 12 auto
 
 **Note - newcol is deprecated**
 
-Older columns (`::newcol`) work, but they are deprecated. Column and row directives should be used instead.  
+Older columns (`::newcol`) work, but they are deprecated. Column and row directives should be used instead.
+
+## 18. Tabs
+
+Directive for adding tabbed content. It means that we can use tabs to separate content into different panels , allowing
```
Mankro, line 1022: This intro could be rephrased, for example: 
 `aplus-tabs` is a directive for adding tabbed content. Tabs separate content into different panels so that one panel is displayed at a time.

Whitespace typo around the comma (but I think the whole comma should be removed): `panels , allowing`

directives/tabs.py
```diff
@@ -0,0 +1,126 @@
...
+        copy_asset(local_path_js, os.path.join(app.outdir, '_static', 'js'))
+        logger.info('done')
+
+
+def setup(app):
+
+    app.add_config_value('include_tab_css', True, 'html')
+    app.add_config_value('include_tab_js', True, 'html')
+
+    app.add_directive('aplus-tabs',  NavTabDirective)
```
Mankro, line 122: Why is the directive named `aplus-tabs`? It is not really tied to A+ at all, is it? If this directive could be very well used outside A+, then should the name be more generic without any connection to A+?

Mankro (changes requested):
I would still ask for some minor changes.

directives/tabs.py
jaguarfi: I will use the `rst-tabs` name instead

jaguarfi:
I  force pushed the changes because the requested changes were rather small

PR #88, opened 2020-05-19
Change the theme's CSS and fix issue #87

### LINKED ISSUE #87: Highlight.js background color and color CSS properties are owewitten by bootstrap
When the `pygments_style` option is changed in the theme settings, the CSS properties, background-color and color,  do not reflect the `pygment_style` value. It seems that bootstrap overwrites the style of the HTML `<pre>` tag and therefore, it overwrites the class highlight.

highlight_issue_01

highlight_issue_02

jaguarfi:
# Description

1. Format the css file of the Aplus theme with Prettier.
2. Fix issue #87: Conflict between the highlight.js styles and Bootstrap
3. Add tip and hint admonitions to the theme's CSS file
4. Change the colour of the download button. The current colour does not match
the theme. The new color `#2491EE` is more appealing, and keep a decent contrast.
However, the links in general should be improved in the future. 
5. Change style of the links inside the `styled-topic` and adjust the alignment of the 
topic and description of the `styled-topic`
6. Set width of the `code-block` to `100%` when it includes the `linenos` option

# Testing
1. The styles were not affected at this point, and everything was working as usual.

2.  Fix issue #87. It was tested manually. The `emacs`, `monokai`, `default`, `vim` and `perldoc` pygments styles were tested in the aplus-manual.
**emacs**
image
**monokai**
image
**default**
image
**vim**
image
**perldoc**
image

3. `hint` and `tip` style
image

4. `download` button/link
image

5. styled-topic
image

6. `code-block` with the `linenos` option now span the same width as the text in the document.
image

----
Some styles in some courses may be affected, but the impact should be minimal.

The documentation does not need to be updated.

# Is it [Done](https://wiki.aalto.fi/display/EDIT/Definition+of+Done)?

-- [ ] I (developer) have created unit tests and the tests pass
-- [ ] I (developer) have created functional tests (Selenium tests) if applicable
-- [X] I (developer) have tested the changes manually
-- [ ] Reviewer has finished the code review
-- [ ] After the review, the developer has made changes accordingly
-- [ ] Customer/Teacher has accepted the implementation of the feature

### DIFF
--- a/theme/aplus/static/default.css
+++ b/theme/aplus/static/default.css
@@ -394,6 +394,8 @@ acronym {
 /* -- code displays --------------------------------------------------------- */
 
 pre {
+  color: inherit;
+  background-color: inherit;
   overflow: auto;
   overflow-y: hidden; /* fixes display issues on Chrome browsers */
 }
--- a/theme/aplus/static/default.css
+++ b/theme/aplus/static/default.css
@@ -41,25 +41,32 @@ special components. Generic and body-level rules come from the A+ system. */
 }
 
 /* Specialized admonitions use different colours */
-.admonition.success, .admonition.meta {
+.admonition.success,
+.admonition.meta {
   color: #3c763d;
   background-color: #dff0d8;
   border-color: #d6e9c6;
 }
 
-.admonition.info, .admonition.note, .admonition.seealso {
+.admonition.info,
+.admonition.note,
+.admonition.seealso {
   color: #31708f;
   background-color: #d9edf7;
   border-color: #bce8f1;
 }
 
-.admonition.warning, .admonition.caution, .admonition.attention, .admonition.admonition-todo {
+.admonition.warning,
+.admonition.caution,
+.admonition.attention,
+.admonition.admonition-todo {
   color: #8a6d3b;
   background-color: #fcf8e3;
   border-color: #faebcc;
 }
 
-.admonition.danger, .admonition.error {
+.admonition.danger,
+.admonition.error {
   color: #a94442;
   background-color: #f2dede;
   border-color: #ebccd1;
@@ -71,20 +78,22 @@ special components. Generic and body-level rules come from the A+ system. */
   border-color: #d7d7d7;
 }
 
-
 /** Add large part of basic css **/
 
 /* -- general body styles --------------------------------------------------- */
 
-div.body p, div.body dd, div.body li, div.body blockquote {
-    -moz-hyphens: auto;
-    -ms-hyphens: auto;
-    -webkit-hyphens: auto;
-    hyphens: auto;
+div.body p,
+div.body dd,
+div.body li,
+div.body blockquote {
+  -moz-hyphens: auto;
+  -ms-hyphens: auto;
+  -webkit-hyphens: auto;
+  hyphens: auto;
 }
 
 a.headerlink {
-    visibility: hidden;
+  visibility: hidden;
 }
 
 h1:hover > a.headerlink,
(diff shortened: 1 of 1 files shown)


raphendyr:
Was this PR automatically linked to the issue or did you add this PR to the issue manually? Seing there is two of those events, I guess Github did ignore the `issue` word between keyword and the issue number.

Any case, I would prefer that everyone would use full strings in the commit title and the body. Person who reads the commit message has no idea what issue number NN is. For example, if we look this raw patch (basically what one could read in the command line), it looks like:

```
From 78c35b9b963f4de6beb6156dddf849d1e75a4add Mon Sep 17 00:00:00 2001
From: Jhosimar Aguacia Fisco <jhosimar.aguaciafisco@aalto.fi>
Date: Tue, 19 May 2020 18:20:44 +0300
Subject: [PATCH] Change the theme's CSS and fix issue #87

* Fix issue #87: Conflict between the highlight.js styles and Bootstrap
* Change the colour of the download button. The current colour does not match
the theme. The new color `#2491EE` is more appealing
* Add tip and hint admonitions to the theme's CSS file
* Set width of the `code-block` to `100%` when it includes the `linenos` option
* Change style of the links inside the `styled-topic`
* Adjust the alignment of the topic and description of the `styled-topic`
```

I would create two commits out of this:

```
Subject: [PATCH] Fix conflict between the highlight.js styles and Bootstrap

Fixes https://github.com/Aalto-LeTech/a-plus-rst-tools/issues/87
```

and

```
Subject: [PATCH] Fix small style issues with Aplus style

* Change the colour of the download button. The current colour does not match
the theme. The new color `#2491EE` is more appealing
* Add tip and hint admonitions to the theme's CSS file
* Set width of the `code-block` to `100%` when it includes the `linenos` option
* Change style of the links inside the `styled-topic`
* Adjust the alignment of the topic and description of the `styled-topic`
```

It's possible the second commit still includes many different changes, but I guess it's fine.

For reference, Coala project seems to have nice style commit message guide: http://api.coala.io/en/latest/Developers/Writing_Good_Commits.html . Sadly,there seems to be no erefences or reasoning why urls are nicer than issue numbers. Allthough, issue numbers seem to be preferred by huge project. I personally prefer full URLs as then you are able to click the link from the command line, instead of searching the repo -> issues -> issue number. It's important to remember that commit messages are often read via command line (e.g. git blame).

jaguarfi:
@raphendyr, it  sounds reasonable. I will split this PR in two commits, and I will refer to the issue using the full url.

raphendyr:
About how Github handles automatic linking: https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

jaguarfi:
@raphendyr I split the commits. Is it good now?

theme/aplus/static/default.css
```diff
@@ -368,10 +368,12 @@ abbr, acronym {
...
     cursor: help;
 }
-
+  
 /* -- code displays --------------------------------------------------------- */
-
+  
 pre {
+    color: inherit;
+    background-color: inherit;
```
raphendyr: So, for the "Fix conflict between the highlight.js styles and Bootstrap" these are the only two required lines?

raphendyr:
Yeah. Scope of the commits are now good. Can be merged.

In the future, I recommend you first do the style changes (fix indentation), and then add the commits fixing concrete issues. Also avoid adding unnesesary spaces (`git add -p` is handy to pick just what you need). As said, this PR is good as is. Time would be wasted on doing these for the PR at this point.

theme/aplus/static/default.css
jaguarfi: Yes, the issue was focus on background color and font color of the highlighted text.

jaguarfi:
Where do I find information about `git add -p`? I could not find much information about it

Mankro:
> Where do I find information about `git add -p`? I could not find much information about it

https://git-scm.com/docs/git-add#Documentation/git-add.txt---patch

jaguarfi:
I can recommend the following links. I found those more explanatory.http://codefoster.com/addpatch/

jaguarfi:
I make this PR a draft, because I am testing the ` git rebase -i` and `git add -p` wit it. Anyway, I will apply the last changes requested by @raphendyr (two commits, remove indentation spaces) and it should be ready to go.

raphendyr:
> I make this PR a draft

Cool, I think it was not possible to change PR back to draft few weeks ago.. Nice to know!

jaguarfi:
Changing the repository of organization workes smoothly. I was able to push some changes without major problems. I think this PR is ready. you can take a look and comment on which commits should be removed or squashed.

jaguarfi:
Changing the repository of organization workes smoothly. I was able to push some changes without problems. I think this PR is ready. you can take a look and comment on which commits should be removed or squashed.

theme/aplus/static/default.css
```diff
@@ -97,467 +108,528 @@ dt:hover > a.headerlink,
...
-    border-top: 2px solid #3c763d;
-    border-bottom: 2px solid #3c763d;
-    background: #ebf5e7;
+  border: 1px solid #e7f2fa;
+  border-top: 2px solid #3c763d;
+  border-bottom: 2px solid #3c763d;
+  background: #ebf5e7;
+}
+
+/* The current version of A+ use Bootstrap 3. The following classes will emulate
```
raphendyr, line 621: Should this be actually implemented in A+? After all, they edit the non course context (i.e area out of `#course-content`). To be fair, all css here should be written so that it only affects content below course area and not the course menu for example.

theme/aplus/static/default.css
```diff
@@ -97,467 +108,528 @@ dt:hover > a.headerlink,
...
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
+    "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
+    "Segoe UI Symbol";
+  /* font-size:  1rem; */
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 1.6;
+  color: #212529;
```
raphendyr, line 632: So, this is a bit more "black" and tinted to blue than the one in bootstrap 3?

jaguarfi:
I was trying to find some courses that use the default.css styles, but it is taking to much time and I do not find anything. Anycase, I think this is not a problem, because these styles can be easily overwritten by the order in which the files are called. Therefore, if the teachers add files to the `layout.html` those will be called after the `default.css` and the styles they define will be used. 

Another point that I would like to highlight is that is not my intention to emulate the Bootstrap 4 fonts. My idea is not about taking the fonts from a newer and more cooler version of Bootstrap, instead this changes are more related to the fact that having a  bigger and modern fonts with specific colors and justification of the text will help to increase the readability of the course content. 
I think the changes related to Boostrap 4 can be renamed with something like `Improbe readability` or similar, but it is my believe that these changes are still relevant for the aplus-theme and will give the user a better experience without the need of zooming in.

theme/aplus/static/default.css
jaguarfi: yes, the color increase the contrast from 12 to 15. Nothing expecial here but it is an small improvement :)

Mankro:
That's alright. There is still the question that should we define any `body` level rules in the a-plus-rst-tools or should they be defined in A+. Since it affects the whole `body`, it is no longer related to the content produced with rst-tools and Sphinx. It affects the A+ menus and everything else in the page.

theme/aplus/static/default.css
jaguarfi: It is true, but if we talk about sphinx themes, those themes are intended to change the whole UI of the documentation. So I am confused now. I am not sure if the intention of having the aplus-theme here is to change the whole UI as the sphinx themes do, or the aplus-theme is only for styling a-plus-rst-tools elements. In that case, this is not a theme in the whole sense of the word. Could you clarify this idea a bit better?

jaguarfi:
@Mankro https://github.com/apluslms/a-plus-rst-tools/pull/88#discussion_r448909266

jaguarfi:
Just as a visual example, the current font size and distribution looks like this:
old-font
While, the text with the small fixes looks like this:
new-font

The changes are not abrupt at all, but they actually distribute the text more evenly and would help the student read better, especially those who may lose track of the line they are reading (a little bit my case :sweat_smile: ).

Mankro:
The screenshot is nice and I'm not against improving the readability of the text, but these styles are not supposed to affect the menu. The course menu in the left does not need the increased line spacing that shows in your screenshot too.

theme/aplus/static/default.css
Mankro: We don't use Sphinx in the conventional way since we are embedding those HTML pages in another system. I don't think our default style in the a-plus-rst-tools should affect the A+ platform itself, like the menus. A-plus-rst-tools should focus on the course contents. If some course has styles that affect the A+ menus too, they can do that, but it is not the default behaviour. If we want to change the styles of the whole platform for everyone, we can do it in A+, not in a-plus-rst-tools.

raphendyr: The course specific CSS shouldn't ever edit A+ frame, only the course content area. In addition, they shouldn't edit A+ elements, like exercise frames, by default. Basically, if we fix some A+ outlooks in the rst-tools, then courses which don't use that would still use the basic styles from A+, which at least makes the courses look different.

I think the "aplus theme" was originally done to make the static html files look-a-like of A+ installation, as it was part of the development workflow. Now that the local files shouldn't normally be checked, we could easily strip away all look-a-like parts (like top and side navigation). Thus, the concept of theme would become simpler.

However, themes would be nice way to support thinks like O1 material, where the material is very different from the default stuff.

theme/aplus/static/default.css
raphendyr: And thus should be done in A+, not here...

jaguarfi:
I will remove the las commit related to the CSS applied to the `body` :). I think the others are acceptable and were already discussed. I will update the PR and add the description about the indentantion.

raphendyr:
> I will remove the las commit related to the CSS applied to the body :). I think the others are acceptable and were already discussed. I will update the PR and add the description about the indentantion.

Can you add those as an issue to A+ and link it to SASS stuff. I think it could be done as part of that.

jaguarfi:
I added this new issue to A+ https://github.com/apluslms/a-plus/issues/609

Issue #68, opened 2020-03-04
Course-specific configuration/information is scattered in different places


jaguarfi:
Course-specific configuration/information is scattered in different places (e.g., the course description and the category settings) and is very much Sphinx-specific (e.g., the content in conf.py). It would be a better option to use a single YAML file to configure the course and generate the other files (including the conf.py file) from that.

**Side note:** This problem is addressed in *Ariel*

Mankro:
Fixing this would be more suitable for a new tool built from scratch (like Ariel) since it changes the course interface and affects the core of the tool.

Mankro:
The old idea for the project named "Ariel" has been discontinued. No implementation was ever started. New ideas for solutions are welcome.

PR #92, opened 2020-08-27
Set module dates from conf.py in aplusmeta directive

atilante:
# Description

Allow setting module dates in the aplusmeta Sphinx directive using a new conf.py variable "aplusmeta_substitutions". See the README.md file at "The meta directive..." for detailed description.

In addition, the aplusmeta directive has now a method validate_time() which raises a SphinxError if the time format is invalid. The purpose of this method is to help a course material developer by:
 (1) causing a RST compilation error at the correct line where the error occurred;
 (2) having detailed, corrective feedback in the compilation error.

This tweak was developed for Aalto University courses CS-A1141/-A1143 Data Structures and Algorithms Y (DSA Y). The reason was that these courses have two course instances and total three language versions in the same git repository. I (the developer) and lecturer Ari Korhonen <ari.korhonen@aalto.fi> has tested and approved the changes. I created the pull request because it might be useful for other courses: all module opening and closing times can be set at the same configuration file. I have attached a copy of the conf.py of DSA Y, see lines 64-114 for an example.

Sidenote 1: if you look at the attached conf.py, DSA Y also has a modified build.sh which sets the value of variable master_doc to either 'index_1141' or 'index_1143' depending on which course is compiled. The related discussion is at aplusguru@cs.aalto.fi email chain with topic " [Aplusguru #16654] CS-A1141/CS-A1143 merge". The tweak is independent of this a-plus-rst-tools pull request.

Sidenote 2: I tried to manually import https://github.com/apluslms/a-plus-rst-tools/commit/c8bbe542d22246ff882e2cea0d10ffe65d071293 into this pull request for easier merging.

# Testing

Unit tests: It seems that A+ RST tools has no unit tests.

Manual testing:
1. Edit conf.py. Set a "aplusmeta_substitutions" variable in conf.py. (See README.md.)
2. Edit an index.rst file of an A+ learning module (chapter). Apply corresponding substitutions. (See README.md.)
3. Recompile your course locally with ./docker-compile.sh .
4. Start A+ locally with ./docker-up.sh .
5. Inspect changes to dates by web browser at the usual http://localhost:8000 .

What could be affected by this changes: setting options of aplusmeta directive directly in an index.rst file should still work. A possible future issue is related to #68 : should Ariel support the feature of this pull request?

Yes. README.md is updated beginning from "The meta directive does not have any content..."

# Is it [Done](https://wiki.aalto.fi/display/EDIT/Definition+of+Done)?

- [ ] I (developer) have created unit tests and the tests pass
- [ ] I (developer) have created functional tests (Selenium tests) if applicable
- [x] I (developer) have tested the changes manually
- [x] Customer/Teacher has accepted the implementation of the feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,36 @@ Module 1 - Introduction
   :close-time: 2020-09-30 14:00
 ```
 
+Alternatively, one can also define these options in the conf.py file of the
+course in the following way.
+
+1. Add the `aplusmeta_substitutions` variable in the conf.py file.
+
+```python
+    aplusmeta_substitutions = {
+        'open01': '2020-09-10 10:01'
+    }
+```
+
+This variable is a dictionary where keys are strings and values are dates
+in the usual format (see above). In the example above, you have defined a
+shortcut text "open01" for date "2020-09-10 10:01".
+
+2. Use the shortcut texts in the aplusmeta directive in a module index.rst:
+
+```rst
+    .. aplusmeta::
+      :open-time: open01
+      :close-time: 2019-09-20 12:00
+      :late-time: 2019-12-20 09:00
+```
+
+3. When the course is compiled, the aplusmeta directive looks for substitution
+strings in the `aplusmeta_substitutions` dictionary. The substitutions can be
+named freely as long as they are not RST markup (e.g. '|open01|' will not
+work). The substitutions can be used with any option of the meta directive.
+
+
 ### 6. Active element input
 
 This creates an input field for active element.
--- a/directives/meta.py
+++ b/directives/meta.py
@@ -1,8 +1,9 @@
+import re
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
+from sphinx.errors import SphinxError
 
 from aplus_nodes import aplusmeta
-
 
 class AplusMeta(Directive):
     ''' Injects document meta data for A+ configuration. '''
@@ -22,5 +23,94 @@ class AplusMeta(Directive):
         'introduction': directives.unchanged, # module introduction HTML
     }
 
+    # Valid date formats are the same as in function parse_date() in
+    # toc_config.py:
+    # 1. 'YYYY-MM-DD [hh[:mm[:ss]]]'
+    # 2. 'DD.MM.YYYY [hh[:mm[:ss]]]'
+    date_format = re.compile(
+        r"^(\d\d\d\d-\d\d-\d\d|\d\d.\d\d.\d\d\d\d)" # YYYY-MM-DD or DD.MM.YYYY
+        "( \d\d(:\d\d(:\d\d)?)?)?$")                # [hh[:mm[:ss]]]
+
+    # Keys in option_spec which require a date format
+    date_format_required = {
+        'open-time',
+        'read-open-time',
+        'close-time',
+        'late-time',
+    }
+
     def run(self):
+        env = self.state.document.settings.env
+        substitutions = env.config.aplusmeta_substitutions
+
+        # Substitute values of options if a corresponding string is found in
+        # the configuration variable aplusmeta_substitutions (set in conf.py).
+        # Example:
+        #     self.options['open-time'] == 'open01'
+        #     aplusmeta_substitutions['open01'] == '2020-01-03 12:00'
+        #     # Result
+        #     modified_options['open-time'] = '2020-01-03 12:00'
+        #
+        # See the section "5. Meta (exercise round settings)" in README.md.
+        for opt, value in self.options.items():
+            old_value = None
+            if value in substitutions:
+                old_value = value
+                self.options[opt] = substitutions[value]
+            if opt in AplusMeta.date_format_required:
+                self.validate_time(opt, self.options[opt], old_value)
+
         return [aplusmeta(options=self.options)]
+
+    def validate_time(self, opt, value, old_value):
+        ''' Validates the time of given option-value pair.
+
+        Parameters:
+        opt:       key of option (one of self.option_spec)
+        value:     value of option (possibly after substitution)
+        old_value: value of option before substitution, or None if substitution
+                   was not used
+
+        Raises a SphinxError if value is invalid.
+        '''
+        if AplusMeta.date_format.match(value):
+            return
+
+        # Raise a SphinxError with a helpful message.
+        source, line = self.state_machine.get_source_and_line(self.lineno)
+
+        msg1 = (
+          "{file}, line {line}, directive aplusmeta:\n"
+          "option '{name}' has a value '{value}' which is not a date or a substitution.\n"
+          "'{value}' should be one of:\n"
+          "1. A date in the format 'YYYY-MM-DD [hh[:mm[:ss]]]', e.g., '2020-01-16 16:00'\n"
+          "2. A date in the format 'DD.MM.YYYY [hh[:mm[:ss]]]', e.g., '16.01.2020 16:00'\n"
+          "3. A substitution text in variable aplusmeta_substitutions in the file\n"
+          "   conf.py. See the A-plus RST tools file README.md > Meta (exercise round \n"
+          "   settings) for reference."
+        )
+        msg2 = (
+          "{file}, line {line}, directive aplusmeta:\n"
+          "option '{name}' has the value '{old_value}' "
+          "which substitutes for value '{value}'. This was not recognised as a date.\n"
+          "Please check the variable `aplusmeta_substitutions` in the file conf.py.\n"
+          "Set the value '{value}' to either:\n"
+          "1. A date in the format 'YYYY-MM-DD [hh[:mm[:ss]]]', e.g., '2020-01-16 16:00'\n"
+          "2. A date in the format 'DD.MM.YYYY [hh[:mm[:ss]]]', e.g., '16.01.2020 16:00'\n"
+        )
+
+        if old_value is None:
+            # No substitution
+            raise SphinxError(msg1.format(
+                file=source,
+                line=line,
+                name=opt,
+                value=value))
+        else:
+            # Substitution is used
+            raise SphinxError(msg2.format(
+                file=source,
+                line=line,
+                name=opt,
+                value=value,
+                old_value=old_value))
--- a/aplus_setup.py
+++ b/aplus_setup.py
@@ -25,6 +25,7 @@ def setup(app):
     app.add_config_value('submit_title', "{config_title}", 'html')
     app.add_config_value('course_open_date', None, 'html')
     app.add_config_value('course_close_date', None, 'html')
+    app.add_config_value('aplusmeta_substitutions', {}, 'html')
     app.add_config_value('questionnaire_default_submissions', 5, 'html')
     app.add_config_value('program_default_submissions', 10, 'html')
     app.add_config_value('default_min_group_size', 1, 'html')
--- a/conf.py
+++ b/conf.py
@@ -20,6 +20,10 @@ import os
 # -- Aplus configuration --------------------------------------------------
 course_open_date = '2016-01-01'
 course_close_date = '2017-01-01'
+aplusmeta_substitutions = {
+    'open01': '2016-01-02 12:00',
+    'close01': '2016-03-01 12:00'
+}
 questionnaire_default_submissions = 5
 program_default_submissions = 10
 ae_default_submissions = 0


directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
+    # Keys in option_spec which require a date format
+    date_format_required = {
+        'open-time', 'read-open-time', 'close-time', 'late-time'
+    }
+
     def run(self):
-        return [aplusmeta(options=self.options)]
+        env = self.state.document.settings.env
+        aplusmeta_substitutions = env.config.aplusmeta_substitutions
+        modified_options = copy.deepcopy(self.options)
```
raphendyr, line 51: Is this deepcopy required?

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
+        #     self.options['open-time'] == 'open01'
+        #     aplusmeta_substitutions['open01'] == '2020-01-03 12:00'
+        #     # Result
+        #     modified_options['open-time'] = '2020-01-03 12:00'
+        #
+        # See the section "5. Meta (exercise round settings)" in README.md.
+        for opt, value in self.options.items():
+            if opt in AplusMeta.date_format_required:
+                self.validate_time(opt, value, aplusmeta_substitutions)
+            if value in aplusmeta_substitutions:
```
raphendyr, line 65: You should do this substitution first, so you wouldn't need to do it late in the validate_time. If you really need to show different error based if it was substituted or not, then do something like.

```python
modified_options[opt] = new_value = aplusmeta_substitutions.get(value, value)
if opt in self.date_format_required:
  self.validate_time(opt, new_value, value)
```

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
+        return [aplusmeta(options=modified_options)]
+
+    def validate_time(self, opt, value, aplusmeta_substitutions):
+        # Validates the time of given option-value pair. Also checks possible
+        # substitutions in the configuration variable 'aplusmeta_substitutions'.
+        # Raises a SphinxError if value is invalid (even with substitution).
+        source, line = self.state_machine.get_source_and_line(self.lineno)
+        position_text = ("{}, line {}, directive aplusmeta:\n" +
+                         "option '{}' has value '{}' ")
+        date_text = ("1. Date in format 'YYYY-MM-DD [hh[:mm[:ss]]]', e.g. " +
```
raphendyr, line 76: Don't use addition to concatenate strings, it's super slow. Just write string

```python
foo = (
  "fist line of string "
  "where it continues here "
  "and some more here."
)
```

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
+                    "'2020-01-16 16:00'\n" +
+                    "2. Date in format 'DD.MM.YYYY [hh[:mm[:ss]]]', e.g. " +
+                    "'16.01.2020 16:00'\n")
+        is_time = AplusMeta.date_format.match
+        if not is_time(value):
+            if value not in aplusmeta_substitutions:
+                raise SphinxError((position_text +
+                    "which was not recognised.\n" +
+                    "This should be one of:\n" + date_text +
```
raphendyr, line 85: This is slower (most likely) than just writing both template strings separately. This also makes the code harder to read as you need to concatenate the strings in your head. Just write both strings with placeholders, so they are easy to read and fix in the code.

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
+                    "3. A substitution text in aplusmeta_substitutions in conf.py; " +
+                    "see A-plus RST tools README.md > Meta (exercise round settings)"
+                    ).format(source, line, opt, value))
```
raphendyr, line 88: There are so many arguments for the format, that it would make life easier to user keyword arguments. Thus,

```python
msg1 = (
  "{file}, line {line}, directive aplusmeta:\n"
  "option '{name}' has value '{value}' which was not recognised.\n"
  "This should be one of:\n"
  "1. Date in format 'YYYY-MM-DD [hh[:mm[:ss]]]', e.g., '2020-01-16 16:00'\n"
  "2. Date in format 'DD.MM.YYYY [hh[:mm[:ss]]]', e.g., '16.01.2020 16:00'\n"
  "3. A substitution text in aplusmeta_substitutions in conf.py; see A-plus RST tools README.md > Meta (exercise round settings)"
)
msg2 = (
  "{file}, line {line}, directive aplusmeta:\n"
  "option '{name}' has value '{value}'  which substitutes to invalid value '{final_value}'.\n"
  "This should be either:\n"
  "1. Date in format 'YYYY-MM-DD [hh[:mm[:ss]]]', e.g., '2020-01-16 16:00'\n"
  "2. Date in format 'DD.MM.YYYY [hh[:mm[:ss]]]', e.g., '16.01.2020 16:00'\n"
  "Probable cause: incorrect field in the variable `aplusmeta_substitutions` in the file conf.py."
)

msg2.format(
  file=source,
  line=line,
  name=opt,
  value=value,
  final_value=new_value, # if change in another comment is implemented
)
```

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
...
     }
 
+    # Valid date formats:
+    # 1. 'YYYY-MM-DD [hh[:mm[:ss]]]'
+    # 2. 'DD.MM.YYYY [hh[:mm[:ss]]]'
+    date_format = re.compile(r"^(\d\d\d\d-\d\d-\d\d|\d\d.\d\d.\d\d\d\d)( \d\d(:\d\d(:\d\d)?)?)?$")
+
+    # Keys in option_spec which require a date format
+    date_format_required = {
+        'open-time', 'read-open-time', 'close-time', 'late-time'
```
raphendyr, line 45: You should have one per line and don't forget comma from the last line.

directives/meta.py
```diff
@@ -22,5 +35,63 @@ class AplusMeta(Directive):
         'introduction': directives.unchanged, # module introduction HTML
     }
 
+    # Valid date formats:
+    # 1. 'YYYY-MM-DD [hh[:mm[:ss]]]'
+    # 2. 'DD.MM.YYYY [hh[:mm[:ss]]]'
+    date_format = re.compile(r"^(\d\d\d\d-\d\d-\d\d|\d\d.\d\d.\d\d\d\d)( \d\d(:\d\d(:\d\d)?)?)?$")
```
raphendyr, line 41: Finnish would use format `DD.MM.YYYY HH.MM.SS`, if you want to be correct. Though, I don't know if that is accepted somewhere else.

directives/meta.py
```diff
@@ -1,8 +1,21 @@
+import copy, re
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
+from sphinx.errors import SphinxError
 
 from aplus_nodes import aplusmeta
 
+def setup(app):
+    # The AplusMeta directive is already set in aplus_setup.py:setup(app).
+    # However, if one wants to use the `aplusmeta_substitutions` variable in conf.py,
```
raphendyr, line 10: Isn't `aplusmeta_substitutions`already registered in the `aplus_setup`, so adding this file to the extensions, shouldn't have any effect? Basically, Sphinx will call the setup from modules listed in the extensions, hence this empty setup doesn't do anything, thus there is no point of adding it to the extensions. Am I right?

README.md
```diff
@@ -200,7 +200,7 @@ The keys for the A+ course settings are listed below:
 ```
 
 Some fields require a value from specific choices (see also
-[the MOOC-grader documentation about the index.yaml file](https://github.com/Aalto-LeTech/mooc-grader/blob/master/courses/README.md)):
+[the MOOC-grader documentation about the index.yaml file](https://github.com/apluslms/mooc-grader/blob/master/courses/README.md)):
```
raphendyr, line 203: These fixes should be in a separate commit. They are not related to the substitution feature.

README.md
```diff
@@ -580,6 +582,61 @@ The meta directive does not have any content and it accepts the following option
...
+This variable is a dictionary where keys are strings and values are dates
+in the usual format (see above). In the example above, you have defined a
+shortcut text "open01" for date "2020-09-10 10:01".
+
+3. Use the shortcut texts in the aplusmeta directive in a module index.rst:
+
+```rst
+    .. aplusmeta::
+      :open-time: open01
+      :close-time: 2019-09-20 12:00
```
raphendyr, line 632: Not really part of this change, but I think it should be possible to write stuff like `2w` here to mark that the module should be open for 2 weeks. Or `${open02} + 2w` to say that it should close 2 weeks after `open02` variable.

README.md
```diff
@@ -580,6 +582,61 @@ The meta directive does not have any content and it accepts the following option
...
+      :late-time: 2019-12-20 09:00
+```
+
+4. When the course is compiled, the aplusmeta directive looks for substitution
+strings in the `aplusmeta_substitutions` dictionary. The substitutions can be
+named freely as long as they are not RST markup (e.g. '|open01|' will not
```
raphendyr, line 638: Would RST substitutions work to implement this feature? Would they be easier? Could they be defined in the `index.rst` for example (I don't think so, but would be cool)?

README.md
Mankro: It's actually causing a git conflict since the Github links have already been fixed in the upstream.

README.md
Mankro: OK, but let's keep it out of this PR's scope.

README.md
Mankro: RST does not support normal substitutions in the directive options and that is the reason for doing a custom implementation in this PR.

directives/meta.py
Mankro: I get the same impression that this empty `setup` function is meaningless and that there is no need to add `meta` to the `extensions` of `conf.py`. The new variable `aplusmeta_substitutions` shall be part of the core `aplus_setup` extension. It shouldn't do any harm even if somebody doesn't want to use it. I don't see the benefit in hiding it behind a tiny extra extension. Like raphendyr said, the variable is already defined in `aplus_setup.py`, thus this empty `setup` can be removed.

directives/meta.py
Mankro: We should at least accept the date formats that have been previously accepted. Otherwise, some old course might not work with the new format check. At a quick glance, the `toc_config.py` seems to responsible for parsing dates: https://github.com/apluslms/a-plus-rst-tools/blob/c8bbe542d22246ff882e2cea0d10ffe65d071293/toc_config.py#L130

Mankro (changes requested):
A useful enhancement! Overall, this looks good, but there are some details to fix.

directives/meta.py
atilante: No; it is an unnecessary safeguard which I will remove.

directives/meta.py
atilante: Done.

directives/meta.py
atilante: Done.

directives/meta.py
atilante: True; this improves code readability.

directives/meta.py
atilante: Very good, I'll do this.

directives/meta.py
atilante: True, that is consistent with the earlier `option_spec` definition.

directives/meta.py
atilante: raphendyr: True. I think it is a separate discussion whether there is need for more date and time formats. Trak Y has separate RST macros for Finnish date texts such as `|tentti1| replace:: 18.12.2020 klo 09-12`.

Mankro: Yes, I duplicated the date format validation from `parse_date()` in `toc_config.py`. To prevent code duplication, it would be best to use only one date parser function. Currently function `parse_date()` cannot be imported in `directives/meta.py`, because `toc_config.py` has line 7: `import directives.meta`. I propose that `parse_date()` is refactored into a new file `lib/config_parsing.py` from which both `toc_config.py` and `directives/meta.py` can import it. It would be good to document the function properly. My proposal:

**lib/toc_config.py**
`
def parse_date(src, allow_empty=False):
    """Parses a date from natural text into and converts it to standard format.
    
    Parameters:
    src: string containing a date text in one of the formats:
         * 'date'
         * 'date time'
         where:
         * 'date' is either in format 'dd.mm.YYYY' or 'YYYY-mm-dd', where
           'dd' day of month (01..31), 'mm' is number of month (01..12) and
           'YYYY' is the year. For example: '25.09.2020' or '2020-09-25'.
         * 'time' is one of the following:
           'HH'       hour in 24-hour clock
           'HH:MM'    minutes
           'HH:MM:SS' seconds
           fallback: set to 12:00 noon

    allow_empty: if True, allows (src == None)

    Returns:
    Date string in one of the following formats:
    'YYYY-mm-dd HH'
    'YYYY-mm-dd HH:MM'
    'YYYY-mm-dd HH:MM:SS'
    """
    if allow_empty and not src:
        return None

    parts = src.split(u' ', 1)
    d = parts[0] # date
    t = parts[1] if len(parts) > 1 else '' # time

    if re.match(r'^\d\d.\d\d.\d\d\d\d$', d):
        # Date in format DD.MM.YYYY (Date, Month, Year)
        # -> convert to YYYY-MM-DD
        ds = d.split('.')
        d = ds[2] + u'-' + ds[1] + u'-' + ds[0]

    elif re.match(r'^\d\d\d\d-\d\d-\d\d$', d):
        # Date in format YYYY-MM-DD (Year, Month, Date):
        # accepted as it is
        pass
    else:
        raise SphinxError(u'Invalid date ' + d)

    if not re.match(r'^\d\d(:\d\d(:\d\d)?)?$', t):
        t = u'12:00'
    return d + u' ' + t
`

Mankro: Yes, `lib/config_parsing.py` is a good idea. And let's start by using the same date formats that have previously been supported in the directive options.

README.md
atilante: Yes, exactly as Mankro said.

directives/meta.py
atilante: I have tested that both `parse_date` in `toc_config.py` and `date_format` in `meta.py` support the same date formats. The following strings are accepted:
`    '01.01.2020',
    '01.01.2020 10',
    '01.01.2020 10:32',
    '01.01.2020 10:32:15',
    '2020-01-02',
    '2020-01-02 15',
    '2020-01-02 15:12',
    '2020-01-02 15:12:11'`

An empty string is not accepted in those aplusmeta directive options that require a date. Reference: `toc_config.py` lines 261-269.
https://github.com/apluslms/a-plus-rst-tools/blob/c8bbe542d22246ff882e2cea0d10ffe65d071293/toc_config.py#L261
I will add a comment on this in the regexp variable `date_format` in `meta.py` and also improve the readability of the regexp by splitting it on several lines with comments.

directives/meta.py
atilante: True, I re-checked that `meta` can be removed from `extensions` in `conf.py`.

README.md
atilante: Good to know this didn't help merging pull requests. I will revert to README.md to d2dd81f and re-apply my changes there.

atilante:
First, notice that this PR isn't urgent; the Trak Y course is already in production using the A+ RST Tools version from my Github repo.

I have now resolved the conflict with README.md and made modifications according to all previous discussion.

Mankro:
@atilante Thank you, the new version looks really good! The code is clean and the README is clear. I trust that you have tested this carefully. I could now merge this into the master branch, but I would first ask you to clean up the git history. You have merged the master branch into your topic branch, but I would prefer that the topic branch is rebased onto the master branch without any merge commits. You should also squash your commits into a single commit since this is a small change. Rebasing may require you to solve the git conflicts in the README again, but that is not hard to do. Rebasing and squashing modifies the history so that you will need to force-push the new version into your topic branch in the fork.

If rebasing is new to you, make a backup first since rebasing can destroy the whole repo when you don't understand what is happening.

Mankro (approved):

atilante:
I have tested that the aplusmeta directive still accepts dates written directly to an RST file.
Tested options: `read-open-time`, `open-time`, `close-time`, `late-time`.
Tested formats:
`YYYY-MM-DD`
`YYYY-MM-DD hh`
`YYYY-MM-DD hh:mm`
`YYYY-MM-DD hh:mm:ss`
`DD.MM.YYYY`
`DD.MM.YYYY hh`
`DD.MM.YYYY hh:mm`
`DD.MM.YYYY hh:mm:ss`
DD = day of month (01-31), MM = number of month (01-12), YYYY = year. hh, mm, ss are hours, minutes, and seconds, respectively.

The error messages related to substitutions have been tested as well.

PR #125, opened 2021-04-27
Fix crash with trilingual courses

skulonen:
# Description

**What?**

Fix crash with trilingual courses.

**Why?**

Trilingual courses caused mooc-grader to crash while starting up. This was caused by a-plus-rst-tools, which generated an invalid YAML structure for these courses.

I made a pull request despite having the privileges to push to a-plus-rst-tools/master, because I haven't touched this repository before and there might be something I hadn't considered.

**How?**

`toc_languages.py` had two issues with multilingual courses.

1) `has_identical_dict_keys` returned `False` when one of the dicts already contained localized values, i.e. a key had a `|i18n` suffix. Because of this, the last branch of `join_exercise_values` would get executed, causing an extra level of localization nesting. This wasn't a problem with bilingual courses because during the first loop (when joining English and Finnish dicts) neither of the original dicts had localized values.

2) `join_keys` should return the common part of the two keys, without the language id suffix. However, this didn't work when the language id was not at the end, e.g. `chapter2_fi_linkquiz`. The new version finds the language ids properly even in the middle of the string.

Fixes apluslms/mooc-grader#96

# Testing

**Remember to add or update unit tests for new features and changes.**

* How to [test your changes in A-plus](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations)
* How to [test accessibility](https://wiki.aalto.fi/display/EDIT/How+to+check+the+accessibility+of+pull+requests)

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that `trilingual`branch of `test-course` now starts up without crashing and localized strings appear properly in all languages. Tested that the course JSON can be exported using the UI without errors, and the structure is like the one generated from the `multilang` branch. Tested the new `join_keys` implementation with many cases, such as "test_final_fi", where the "_fi" part of "_final" could cause problems in an incorrect implementation.

**Did you test the changes in**

- [ ] Chrome
- [x] Firefox
- [ ] This pull request cannot be tested in the browser.

**Think of what is affected by these changes and could become broken**

# Translation

- [ ] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [x] Did you follow our [style guides](https://apluslms.github.io/contribute/styleguides/)?

- [ ] documents inside the doc directory.
- [ ] README.md.
- [ ] Aplus Manual.
- [ ] Other documentation (mention below which documentation).

# Is it Done?

- [ ] Customer/Teacher has accepted the implementation of the feature

### DIFF
--- a/lib/toc_languages.py
+++ b/lib/toc_languages.py
@@ -315,10 +315,24 @@ def join_values(lang1, val1, lang2, val2):
 
 
 def has_identical_dict_keys(d1, d2):
-    return (
-        type(d1) == type(d2) == dict
-        and len(set(d1.keys()) ^ set(d2.keys())) == 0
-    )
+    if not (type(d1) == type(d2) == dict):
+        return False
+
+    # Collect the keys of both dicts into sets and compare their differences.
+    # Keys with and without the i18n suffix are considered identical.
+    keys1 = set()
+    for k in d1:
+        if k.endswith('|i18n'):
+            keys1.add(k[:-5])
+        else:
+            keys1.add(k)
+    keys2 = set()
+    for k in d2:
+        if k.endswith('|i18n'):
+            keys2.add(k[:-5])
+        else:
+            keys2.add(k)
+    return len(keys1 ^ keys2) == 0
 
 
 def has_identical_len_and_dict_keys(l1, l2):


Mankro (commented):
Great observations! Did you test this with monolingual and two-lingual courses (test-course branches master and multilingual) in addition to the trilingual case?

You mentioned the language code showing up in the middle of `chapter2_fi_linkquiz`. The `_build/yaml` directory has always had these files in two-lingual courses. Now I wonder if your new changes might interfere with something. I really don't know yet, but I would ask you to test more cases first.

skulonen:
Yes, I've confirmed that it works with monolingual and bilingual courses.

> The _build/yaml directory has always had these files in two-lingual courses. Now I wonder if your new changes might interfere with something.

Good point. I recently fixed apluslms/a-plus#719, and now I feel like it could cause some nasty issues if two versions of A+ generate different keys for modules and exercises. I just tested, and found out that the `join_keys` fix is not actually necessary for the `trilingual` branch to work. So, we could leave that part out of the PR to be safe.

The reason I originally made the fix to `join_keys` was because I noticed there were some yaml files in the output directory with super odd names, like `links1_chapter1_p` and `links1_chapter2_i`, and that many of the files were redundant, and that the order in which the languages are loaded causes differences in the file names. This lead me to the `join_keys` function, which turned out to be very buggy. But I didn't realize that we may not be able to change the key generation logic.

Mankro:
Yes indeed, changing the keys would cause issues in any existing course when they upgrade the a-plus-rst-tools submodule. Let's not modify keys since it is not necessary to support trilingual courses.

skulonen:
I reverted the `join_keys` fix and tested that the trilingual course still works.

Mankro (approved):
Great fix! Does this version have the problem with super odd filenames like `links1_chapter1_p` and `links1_chapter2_i`? If the odd files exist, then do they affect anything or are they redundant? I guess that in bilingual courses, the different language files like `chapter2_fi_linkquiz` are mostly redundant, but they are maybe needed until the combined version `chapter2_linkquiz` is written to the disk. I haven't studied the code so I am only guessing. Maybe someday we could clean up and remove the redundant files, but it is not important right now.

Would you squash the commits and update the git commit message (and force-push)? Squashing removes the reverted changes in the other function (`join_keys`) from the history.

skulonen:
> Does this version have the problem with super odd filenames like links1_chapter1_p and links1_chapter2_i?

Yes, the odd files are caused by the buggy `join_keys` function, and because I reverted the fix, the still exist. As far as I can tell from testing the trilingual branch, they don't cause any problems. But they seem to be redundant because they don't appear if the join_keys fix is in place (and the course still works with the fix).

Sure, I'll squash them.

skulonen:
The commits are now squashed and the commit message updated.

PR #156, opened 2023-01-03
Add yaml dependency to rst files

lainets:
# Description

Adds yaml files as dependencies to the rst files that reference them, allowing incremental sphinx builds.


lainets:
Note that the build output directory needs to be specified in the `exclude_patterns` field of the apps.meta file, otherwise the git clean command in Git Manager will remove the files and the build will be done from scratch. The build command for the course needs to also be modified so that touchrst isn't run anymore.

Mankro (approved):
Great! I have been testing this and it seems to work well.

Mankro:
I force-pushed because I adjusted the git commit message first.

PR #166, opened 2023-06-08
Add category configuring, and default accept unofficial to true for active elements

### LINKED ISSUE #162: Active elements category cannot be changed
The active elements category is hard coded to "active elements". https://github.com/apluslms/a-plus-rst-tools/blob/1a7d6e007f5127e1d68c8ddf0a5fa0cec3845f10/directives/ae_output.py#L109

Fixing this may affect a course if it tries to change the category and nobody noticed that it doesn't work.

### LINKED ISSUE #164: Categories are difficult to configure
Categories cannot be configured easily. E.g. setting `accept_unofficial_submits` is difficult. `append_content` in conf.py can be used but it is a bit cumbersome.

### LINKED ISSUE #165: Default active elements to allow unofficial submissions
This allows submission after the deadline. Needed for https://github.com/apluslms/a-plus/issues/487

#164 should be fixed first so that the behaviour can be easily changed if the teacher wants to.

It should probably not be enabled by default if max submissions is specfied. As `accept_unofficial_submits` is a category specific setting, this would either mean disabling unofficial submission if even one active element exercise has max submissions specified or two different default categories for active elements depending on whether max submission is specified or not.

lainets:
# Description

**What?**

Add easier category configuring, and accept unofficial submissions by default to active elements exercises without max submissions.

**Why?**

To close #165.

**How?**

Active elements with max submissions are put in a separate category "active elements with max submissions" that doesn't accept unofficial submisions by default. 

Closes #162
Closes #164
Closes #165 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tried changing the category options in conf.py, and checked that the changes happen in A+. Also checked that "active elements" category accepts unofficial submission by default and that it can be overridden in conf.py.

### DIFF
--- a/directives/ae_output.py
+++ b/directives/ae_output.py
@@ -109,6 +109,8 @@ class ActiveElementOutput(ConfigurableExercise):
             'category': 'active elements',
             'max_submissions': self.options.get('submissions', data.get('max_submissions', env.config.ae_default_submissions)),
         })
+        data["category"] = "active elements" if data["max_submissions"] == 0 else "active elements with max submissions"
+
         data.setdefault('status', self.options.get('status', 'unlisted'))
 
         self.apply_override(data, category)
--- a/directives/ae_output.py
+++ b/directives/ae_output.py
@@ -3,6 +3,8 @@
 Directive that places active element output divs.
 '''
 import os.path
+from typing import Any, Dict
+
 from docutils.parsers.rst import directives
 from docutils import nodes
 from sphinx.errors import SphinxError
@@ -84,6 +86,7 @@ class ActiveElementOutput(ConfigurableExercise):
         key_title = "{} {}".format(translations.get(env, 'exercise'), key)
 
         # Load or create exercise configuration.
+        data: Dict[str, Any]
         if 'config' in self.options:
             path = os.path.join(env.app.srcdir, self.options['config'])
             if not os.path.exists(path):
@@ -106,10 +109,17 @@ class ActiveElementOutput(ConfigurableExercise):
             'title': env.config.submit_title.format(
                 key_title=key_title, config_title=config_title
             ),
-            'category': 'active elements',
             'max_submissions': self.options.get('submissions', data.get('max_submissions', env.config.ae_default_submissions)),
         })
-        data["category"] = "active elements" if data["max_submissions"] == 0 else "active elements with max submissions"
+
+        if "category" in self.options:
+            data["category"] = self.options["category"]
+        elif "category" not in data:
+            data["category"] = (
+                "active elements"
+                if data["max_submissions"] == 0 else
+                "active elements with max submissions"
+            )
 
         data.setdefault('status', self.options.get('status', 'unlisted'))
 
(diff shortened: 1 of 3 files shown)


lainets:
Of note is whether the solution to the max submissions/accept unofficial submissions problem is satisfactory. Does changing the category of active elements exercises with max submissions cause issues to some courses?

toc_config.py
```diff
@@ -395,15 +395,31 @@ def parse_chapter(docname, doc, parent, module_meta):
 
     # Create categories.
     category_names = app.config.category_names
+    for key, name in category_names.items():
+        category = app.config.categories.get(key, {"name": name})
+        if name != category.get("name", name):
```
Mankro, line 400: `category_names` in `conf.py` may set multilingual names. Isn't this comparison a bit suspicious in that case?

```python
category_names = {
    'chapter': {
        'fi': 'Luku',
        'en': 'Chapter',
    },
    'submit': {
        'fi': 'Pisteytetty',
        'en': 'Graded',
    },
    'feedback': {
        'fi': 'Palaute',
        'en': 'Feedback',
    },
}

```

toc_config.py
```diff
@@ -395,15 +395,31 @@ def parse_chapter(docname, doc, parent, module_meta):
...
         key: {
             'name': category_names.get(key, key),
         } for key in category_keys
     }
+
+    for key, category in app.config.categories.items():
+        if key not in categories:
+            raise SphinxError(f"Category with key '{key}' was configured but not used in the material")
+
+        categories[key].update(category)
```
Mankro, line 413: Should this part be moved after the following `for` loop that forces `chapter` and `feedback` categories to `status` `nototal`? If somebody really wanted to set a different status, then they could do it with the new `categories` setting.

Some courses actually want to show feedback points to the students and changing the status of the feedback category would be then useful.

directives/ae_output.py
```diff
@@ -109,6 +109,8 @@ def run(self):
             'category': 'active elements',
             'max_submissions': self.options.get('submissions', data.get('max_submissions', env.config.ae_default_submissions)),
         })
+        data["category"] = "active elements" if data["max_submissions"] == 0 else "active elements with max submissions"
```
Mankro, line 112: Looks like it wasn't possible to change the category of the active element in the RST directive options. But we could now add the possibility. If somebody wants to change the category, then we also do not need to set default setting (`accept_unofficial_submits`) to the new category.

Do you think this is OK?

Mankro (changes requested):
Great, this looks good! I have a couple of comments below.

> Of note is whether the solution to the max submissions/accept unofficial submissions problem is satisfactory. Does changing the category of active elements exercises with max submissions cause issues to some courses?

I think it is good. I don't think changing the category hurts any current user.

directives/ae_output.py
lainets: Yeah, I don't think we want to set `accept_unofficial_submits` on an user specified caregory as we don't really know what the intention is and what else is in that same category. I'll add the possibility of changing the category.

Mankro (approved):
Excellent!

PR #184, opened 2026-03-04
Support Sphinx up to version 7.4.

### LINKED ISSUE #177: Support new Sphinx releases
AFAIK, a-plus-rst-tools supports Sphinx up to version 4, which is years old by now. At the time of writing, the newest Sphinx release is 7.3. (The Sphinx changelog is [here](https://www.sphinx-doc.org/en/master/changes.html).) Supporting newer Sphinx versions soon would be great. 

(Also, I’m not sure if the older Sphinx versions even work with new versions of Python. At least I had some trouble with this when I tried. Presumably folks are gradually moving to new Python versions.)

Some observations based on quick local experiments with my O1 course materials:

* Supporting Sphinx versions up to 6.2.1 looks unproblematic (at least as far as my materials are concerned). There are a few small changes to HTML output that teachers (such as I) who use custom css/js need to check.
* Supporting Sphinx 7.0.0 and up means a tiny change would be needed in conf.py files and the theme template, because [the `style` key has been replaced by `styles`](https://www.sphinx-doc.org/en/master/changes.html).
* Supporting [Sphinx 7.2.0](https://www.sphinx-doc.org/en/master/changes.html#release-7-2-0-released-aug-17-2023) and up requires some more effort. Specifically, what happened when I tried using a-plus-rst-tools with Sphinx 7.2 (and 7.3) was this:
  * `toc_config.py` gives warnings that say something like "Starting from Sphinx 8, it won’t be OK to handle paths as strings — use `Path` objects instead".
  * Moreover, even through those warnings were apparently intended as mere advance warnings about the upcoming Sphinx 8, `toc_config.py` actually crashes with a path-related error (that I didn’t examine more closely).

sayravai:
Fixes #177

# Description

Add support for Sphinx version 7.4. The main changes to RST tools 1.4 from the user perspective are:
1. master_doc is replaced by root_doc in the project settings (this change was made already in Sphinx 4.0).
2. custom stylesheets are now loaded from the "styles" array instead of a "style" string.

Also the Python 3.12+ warnings about regex patterns (SyntaxWarning: invalid escape sequence '\d', SyntaxWarning: invalid escape sequence '\w') are fixed in this PR.

This could become RST tools v1.5 and we could then later make a version 1.6 with support for Sphinx 9.x if deemed necessary.

# Testing

This PR still needs more testing. I have tested that the rst compilation process for A+ manual ([the version fixed for Sphinx 7's stricter rules](https://github.com/apluslms/aplus-manual/pull/75)) succeeds without warnings using Sphinx v7.4.7, but I haven't checked what the resulting HTML looks like in A+. This should be tested, and preferably compiling some actual course also.

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

[ADD A DESCRIPTION ABOUT WHAT YOU TESTED MANUALLY]

**Did you test the changes in**

- [ ] Chrome
- [ ] Firefox
- [ ] This pull request cannot be tested in the browser.

**Think of what is affected by these changes and could become broken**

# Translation

- [ ] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [ ] Did you follow our [style guides](https://apluslms.github.io/contribute/styleguides/)?
- [ ] Did you use Python type hinting in all functions that you added or edited? ([type hints](https://docs.python.org/3/library/typing.html) for function parameters and return values)

- [ ] documents inside the doc directory.
- [ ] README.md.
- [ ] Aplus Manual.
- [ ] Other documentation (mention below which documentation).

# Is it Done?

- [ ] Customer/Teacher has accepted the implementation of the feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,26 @@ though it also supports other markup languages such as markdown via extensions.
 Sphinx itself extends the Docutils RST parser and compiler.
 A-plus RST tools support the following Sphinx versions:
 
-* a-plus-rst-tools v1.4 support Sphinx v4.1.
+* a-plus-rst-tools v1.5 and later support Sphinx v7.x (up to v7.4).
+* a-plus-rst-tools v1.4 supports Sphinx v4.1.
 * a-plus-rst-tools v1.3 and earlier support Sphinx v1.6.
+
+### Migrating an existing course to RST tools 1.5
+
+If you are upgrading from RST tools v1.4 or earlier to v1.5, one change is
+required in your course `conf.py`: the Sphinx setting `master_doc` was renamed
+to `root_doc`. Open `conf.py` and replace the line
+
+    master_doc = 'index'
+
+with
+
+    root_doc = 'index'
+
+Note also that v1.5 (with Sphinx v7) changes how custom stylesheets are passed to the
+theme. If your course uses a custom CSS file via `html_style` in `conf.py`,
+you should verify that the stylesheet still loads correctly after the upgrade.
+The a-plus theme templates have been updated accordingly.
 
 ## Creating a new course
 
@@ -27,7 +45,7 @@ the course using Docker containers (recommended). If you do not use containers,
 you need Python modules Sphinx and yaml in order to compile the RST source into
 HTML and YAML course configuration files.
 
-    pip install sphinx~=4.1.2 PyYAML~=5.4.1
+    pip install "sphinx>=7.0,<8" PyYAML~=6.0
 
 The course is compiled with make (when no containers are used).
 
@@ -53,7 +71,7 @@ In order to build the course without any Docker containers, you install the
 Python modules Sphinx and yaml as well as create the Sphinx configuration file
 conf.py.
 
-    pip install sphinx~=4.1.2 PyYAML~=5.4.1
+    pip install "sphinx>=7.0,<8" PyYAML~=6.0
     sphinx-quickstart
     cp a-plus-rst-tools/conf.py .  # a similar conf.py is available in course-templates as well
 
--- a/theme/aplus/layout.html
+++ b/theme/aplus/layout.html
@@ -40,13 +40,12 @@
     <!--<link rel="stylesheet" href="https://plus.cs.aalto.fi/static/css/main.css">
     <script src="https://plus.cs.aalto.fi/static/js/bootstrap3.min.js"></script>-->
 
-    {%- if style %}
-    {# This is either the stylesheet defined in theme.conf (default.css) or
-       a custom CSS file defined in the course conf.py file (variable html_style).
-       The custom stylesheet may use `@import url("default.css");` in order to
-       include the stylesheet from the theme. #}
-    <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" data-aplus="yes" />
-    {%- endif %}
+    {%- for stylesheet in styles %}
+    {# styles is a list of stylesheets from theme.conf and/or html_style in conf.py.
+       A custom stylesheet may use `@import url("default.css");` in order to
+       include the default stylesheet from the theme. #}
+    <link rel="stylesheet" href="{{ pathto('_static/' + stylesheet, 1) }}" type="text/css" data-aplus="yes" />
+    {%- endfor %}
     <!--<link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" data-aplus="yes" />-->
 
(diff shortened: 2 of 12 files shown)


theme/aplus/layout.html
```diff
@@ -94,8 +93,7 @@
                 </a>
                 <ul class="dropdown-menu">
                   <li role="presentation">
-                    <a href="{{ pathto(master_doc) }}">
-                      Index
+                    <a href="{{ pathto(root_doc) }}">
                     </a>
```
ihalaij1, line 97: Is there a reason "Index" was removed here but not in some of the other places?

theme/aplus_bs3/layout.html
```diff
@@ -94,8 +93,7 @@
                 </a>
                 <ul class="dropdown-menu">
                   <li role="presentation">
-                    <a href="{{ pathto(master_doc) }}">
-                      Index
+                    <a href="{{ pathto(root_doc) }}">
                     </a>
```
ihalaij1, line 97: Is there a reason "Index" was removed here but not in some of the other places? x2

README.md
```diff
@@ -12,9 +12,27 @@ though it also supports other markup languages such as markdown via extensions.
 Sphinx itself extends the Docutils RST parser and compiler.
 A-plus RST tools support the following Sphinx versions:
 
-* a-plus-rst-tools v1.4 support Sphinx v4.1.
+* a-plus-rst-tools v1.5 and later support Sphinx v7.x (up to v7.4).
+* a-plus-rst-tools v1.4 supports Sphinx v4.1.
```
ihalaij1, line 16: We already have existing a-plus-rst-tools versions 1.5, 1.6, 1.7, and 1.8.
The latest version is 2.0.0 with BS5 support, so shouldn't the next version be 2.1.0?

README.md
```diff
@@ -12,9 +12,27 @@ though it also supports other markup languages such as markdown via extensions.
...
+    master_doc = 'index'
+
+with
+
+    root_doc = 'index'
+
+Note also that v1.5 (with Sphinx v7) changes how custom stylesheets are passed to the
+theme. If your course uses a custom CSS file via `html_style` in `conf.py`,
+you should verify that the stylesheet still loads correctly after the upgrade.
+The a-plus theme templates have been updated accordingly.
```
ihalaij1, line 34: Isn't there also another change required for the `rst_prolog` setting in conf.py (mentioned in the aplus-manual PR).

ihalaij1 (changes requested):
Great!

I tested this with the aplus-manual PR and did not notice any obvious issues on the A+ manual materials.

I added some comments.

theme/aplus_bs3/layout.html
sayravai: I don't think it makes sense to create empty links there, so I guess it's a copy-paste or search & replace error or something...

README.md
sayravai: Seems I didn't check the actual versions from GitHub, but just blindly updated the README...

README.md
sayravai: Hmm, doesn't ring a bell. Which PR are we talking about?

ihalaij1: Here, in conf.py:
https://github.com/apluslms/aplus-manual/pull/75/changes#diff-6e03d3d89630aba3403afea3514a3ac5948e70d1ced091012e56446f614da340R73-R76

sayravai: I don't see rst_prolog mentioned in there. Or should I understand what the setting actually does...?

ihalaij1: It's the comment you added that is required so that the first content isn't considered as part of the last directive.

ihalaij1: If you expand the code in the diff view, you see it is part of rst_prolog.

sayravai: Ah, now I see what you mean. Indeed, if someone has copied that conf.py for the course, it needs to be pached there as well. Sorry, I'm really not very familiar with building courses in A+.

ihalaij1: At least Y1 is using the rst_prolog setting in conf.py, but with different directives than A+ manual.

ihalaij1:
I created a PR for compile-rst to update it to use Sphinx 7:
https://github.com/apluslms/compile-rst/pull/8

sayravai:
> I created a PR for compile-rst to update it to use Sphinx 7: [apluslms/compile-rst#8](https://github.com/apluslms/compile-rst/pull/8)

Nice. Debian bullseye support ends this June, so those references could be upgraded (and tested) soon as well.

